PR #10, opened 2021-04-08
Index type hints

jorants:
Adds an additional less verbose way of adding indices using type hints:
```python
from beanie import Indexed

class DocumentTestModelWithIndex(Document):
    test_int: Indexed(int)
    test_list: List[SubDocument]
    test_str: str
```

I'm not sure the current method of implementing Indexed is the best, but it works and seems like a starting point.

### DIFF
--- a/beanie/__init__.py
+++ b/beanie/__init__.py
@@ -1,6 +1,7 @@
 from beanie.cursor import Cursor
 from beanie.documents import Document
 from beanie.general import init_beanie
+from beanie.collection import Indexed
 
 __version__ = "0.3.2"
-__all__ = ["Document", "Cursor", "init_beanie"]
+__all__ = ["Document", "Cursor", "init_beanie", "Indexed"]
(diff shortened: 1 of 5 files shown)


roman-right:
Hi @jorants ,

Nice PR, I like it!
Could you pls add a specific test for this case and rebase it to a single commit?
Also, it would be nice, if you will use pre-commit. Run `pre-commit install` in the project root to enable it. 

Thank you for the contribution!

jorants:
Hi,

No problem, should be fine now. 

To get the tests to work on my machine I had to add `?authSource=admin` to the database connection.
A few words on setting up the development environment in the documentation could be useful at some point (as the project hopefully becomes more popular!).

roman-right (approved):

roman-right:
Yes, you are right. 
I'm working on a massive update now, which (in plan) will include the whole documentation rework. So, this information must be there definitely :-)

PR #11, opened 2021-04-09
Small addition to the Indexed function

jorants:
Small addition to the previous PR, Index now takes an optional argument to allow setting the index type. 
Sorry, i realized this would make sense to add a few minutes after you merged the previous one.

### DIFF
--- a/beanie/collection.py
+++ b/beanie/collection.py
@@ -5,14 +5,14 @@ from pydantic.main import BaseModel
 from pymongo import IndexModel, ASCENDING
 
 
-def Indexed(typ):
+def Indexed(typ, index_type=ASCENDING):
     """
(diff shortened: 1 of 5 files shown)


roman-right:
Hi!
I like this :-) 

Could you please add a specific test with the non-default value (like in the example in the doc)? To have all the cases tested.

jorants:
I have split up the tests and added one for a pymongo.TEXT argument.
To be more compatible with other versions of mongodb it only checks the key value, not the whole index structure.

roman-right (approved):
Looks good

roman-right:
> I have split up the tests and added one for a pymongo.TEXT argument.
> To be more compatible with other versions of mongodb it only checks the key value, not the whole index structure.

Perfect. Looks very neat.
I will merge this guy a little later today or tomorrow.

Issue #12, opened 2021-04-12
Not possible to use custom type for _id field


paul-nameless:
Something like this does't work:
```
class Task(Document):
    id: str = Field(..., alias="_id")
    updated: datetime = Field(..., default_factory=datetime.utcnow)
```

It would be cool to be able to set _id manually.

roman-right:
Hi @paul-nameless ,

`_id` field itself is reserved by MongoDB to be used as a primary key. Typically it uses ObjectId for this. Correct me, if I'm wrong. Do you mean it would be useful to be able to use other data types for this primary key, right? Or are you talking about another use case?

paul-nameless:
To be fear it's not reserved, it's required https://docs.mongodb.com/manual/reference/bson-types/#objectid

_id is often used as non ObjectId type as a "free" unique index, which doesn't require additional index setup, disk space and memory. This is quite crucial for big collections. 

So that would be great to make it more flexible

roman-right:
First of all, thank you :-) I'm happy that it is useful.

I see. Yes, I do agree, it would be better to have this more flexible. This is linked with some internal logic and, unfortunately, it is impossible to replace it on the move fast. But I will put this improvement at the top of the backlog and will pick it up after the current major update.

Thank you.

paul-nameless:
Basically, something like this let autogenerate _id if it's not set and use the given value otherwise.

https://github.com/paul-nameless/beanie/commit/7d4d78c83b86382e40d6476d8e8570cd20cfc540

That's just draft, if you don't see any issues and happy with this, I could continue working on it.

roman-right:
Yes, this will partially work

The idea of this check is to understand if the document was already inserted or not. Bc if it was inserted, it will not be inserted again - the driver will just ignore the operation. It will not update anything and will not even warn, as I remember, about the double insert attempt. This can confuse. That's why there are two methods `create` to insert the document and `replace` for the full update.

There are some ways, how to solve this using custom id types: 
- A document can have an additional flag, like `_is_inserted` to manage these states. It must be hidden (protected). It will not protect from all the possible collisions, but most of them will be avoided. 
- `create()` method can check if the document with this id was already inserted into the database. It will solve all the collisions, but also it will increase the number of calls.
- Probably smth else, I didn't think well about this.

I prefer the first one for now. 

So yes, if you want to develop this feature - you are welcome for sure :-) 
If you want to chat about any details, you can do it in discord. The link is in the project readme

PR #14, opened 2021-04-14
Allow custom type for id field

paul-nameless:
#12

### DIFF
--- a/beanie/documents.py
+++ b/beanie/documents.py
@@ -87,12 +87,12 @@ class Document(BaseModel):
         Create the document in the database
         :return: Document
         """
-        if self.id is not None:
-            raise DocumentAlreadyCreated
+
+        exclude = None if self.id else {"id"}
         result = await self.get_motor_collection().insert_one(
-            self.dict(by_alias=True, exclude={"id"})
+            self.dict(by_alias=True, exclude=exclude)
         )
-        self.id = PydanticObjectId(result.inserted_id)
+        self.id = result.inserted_id
         return self
 
     @classmethod
--- a/beanie/documents.py
+++ b/beanie/documents.py
@@ -1,21 +1,21 @@
 from enum import Enum
-from typing import Optional, List, Type, Union, Tuple
+from typing import List, Optional, Tuple, Type, Union
 
 import pymongo
 from bson import ObjectId
-from motor.motor_asyncio import AsyncIOMotorDatabase, AsyncIOMotorCollection
+from motor.motor_asyncio import AsyncIOMotorCollection, AsyncIOMotorDatabase
 from pydantic import Field, validator
-from pydantic.main import BaseModel
-from pymongo.results import DeleteResult, UpdateResult, InsertOneResult
+from pydantic.main import BaseModel, PrivateAttr
+from pymongo.results import DeleteResult, InsertOneResult, UpdateResult
 
 from beanie.collection import collection_factory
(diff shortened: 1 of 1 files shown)


paul-nameless:
Hello, please have a look and let me know what you think about patch.
I ran tests locally and they passed

beanie/documents.py
```diff
@@ -87,12 +92,19 @@ async def create(self) -> "Document":
...
             raise DocumentAlreadyCreated
+        exclude = None if self.id else {"id"}
         result = await self.get_motor_collection().insert_one(
-            self.dict(by_alias=True, exclude={"id"})
+            self.dict(by_alias=True, exclude=exclude)
         )
-        self.id = PydanticObjectId(result.inserted_id)
+
+        if isinstance(result.inserted_id, ObjectId):
+            self.id = PydanticObjectId(result.inserted_id)
```
roman-right, line 103: It looks like it is better to use annotation class to return not only PydanticObjectId, but all of them. Because in another case it will return not the type, which was annotated, and it will confuse. For example, there is the next class:
```python
class TestCase(Document):
    id: UUID = ...
    smth: str
```

In this case, `create()` method will return object with id of type `str` (because json doesn't have UUID types and presents it as a string). 

You can get the id annotation with python inspect module:

```python
class_signature = signature(TestCase)
id_annotation = class_signature.parameters["id"].annotation
```
But the thing is, it can be Optional, or even Union. I think it is too complicated to code own parser here for this case. So you can use the next trick (full working example):

```python
from inspect import signature
from typing import Optional
from uuid import UUID

from pydantic.main import BaseModel


class TestCase(BaseModel): # BaseModel instead of Document for demo, to not init collection
    id: Optional[UUID] = None
    smth: str = "wer"

t = TestCase()

ID_STRING_FROM_DB = "91f02ed2-a07c-4662-8bd1-2f38bc67f0fc"

t.id = ID_STRING_FROM_DB
print(t.id, type(t)) # Here id will be str, even if type is UUID in the annotation

class_signature = signature(TestCase)
id_annotation = class_signature.parameters["id"].annotation


class IdParser(BaseModel):
    id: id_annotation


t.id = IdParser(id=ID_STRING_FROM_DB).id

print(t.id, type(t.id)) # It is UUID now

```

It is a bit tricky and, probably, slow. I don't think it is too slow and will be a bottleneck, but probably here there is room for improvement.

roman-right (changes requested):
Hi @paul-nameless

Good job!

I wrote a comment in the code.

Also, I found some other things:
- I think you missed one place, where id is used - https://github.com/roman-right/beanie/blob/feature/migrations/beanie/odm/documents.py#L217
- In all the find* methods you don't mark documents as created, it means they will use logic for not created documents.
- Please add tests for all the cases, where custom `id` will be used. I mean document creation, document replacing, probably document updating, and finding the documents. 
- Probably it is better to use _is_inserted instead of _is_created. But this is more a matter of taste, for sure.

roman-right:
The feature is implemented in the PR https://github.com/roman-right/beanie/pull/42

paul-nameless:
Cool, thanks. I will check it out.
Btw, instead of
```
new_id = self.__fields__["id"].type_(new_id)
```
can be done this:
```
pydantic.parse_obj_as(self.__fields__["id"].type_, new_id)
```
I didn't check if there are any benefits but this is how I wanted to do it.

Issue #16, opened 2021-04-16
Project only the fields that are in our model


nzsmith1:
First of all, thanks for creating this project, it has been fantastic to work with.  I have a feature request, which I am happy to implement, I would like the find, find_many, find_all to project our model so that we get back only the data we are interested in.

This helps reduce database load.

An example of where we might use this is when we have a database model which hold a whole bunch of data, but we are loading a summary view which only requires 2 or 3 of these fields.

example:

```python
class AnalyticsLabellingTask(Document):
    reported_by: str = Field(...)
    device_id: str = Field(...)
    priming_start_date: datetime = Field(...)
    event_start_date: datetime = Field(...)
    end_date: datetime = Field(...)
    reviewers: List[str] = Field(...)
    comments: List[Comments] = Field(default_factory=list)
    reviewed_events: List[ReviewedEvents] = Field(default_factory=list)
    water_data: List[TimeSeriesData] = Field(...)
    archived: bool = Field(...)

    class Collection:
        name = "AnalyticsLabellingTasks"

class AnalyticsLabellingTasksSummary(Document):
    device_id: str = Field(...)
    event_start_date: datetime = Field(...)
    end_date: datetime = Field(...)
    reported_by: str = Field(...)

    class Collection:
        name = "AnalyticsLabellingTasks"
```

I am happy to implement this, I would do it by updating the find calls:

```python
# get the field list of interest, this could be generated
# at the __init__ stage and be stored in memory to prevent
# having to build this many times

fields = test.__fields__
projection = {}
for name, field in fields.items():
    if field.alias:
        projection[field.alias] = 1
    else:
        projection[name] = 1

cursor = cls.get_motor_collection().find(filter_query, projection, **kwargs)
```

Let me know your thoughts and I can put together a PR next week

roman-right:
Hello Nicholas,

Thank you for your feedback.
Yes, this is a great idea. It will so good if you will implement this.

I still didn't add development guides. If you will need any help with this (like how to test and etc.) you can ask here or in Discord.

Thank you.

PR #18, opened 2021-04-19
Projections

nzsmith1:
Added a function to build the projections, this function will be called once when we first ask for it.  The overhead is low, but still makes sense to cache this.  This is cached as an "internal" attribute named `_projection`.  We get the attribute by calling `_get_projection()` this returns the projection if it exists or builds it and returns it.  We use pydantic's `__fields__` value to find all of the items in the model we care for.

I added a quick test as well for clarity, it is likely not needed as this line is called during the .findx() tests but thought it was worthwhile for documenting the expected return data.

I also changed the filter args on the findx calls to kwargs again for clarity as to what we are calling into on each call.


nzsmith1:
Link: #16

roman-right:
Awesome! This is a really important feature.
Could you please rebase this into a single commit? Thank you.

nzsmith1:
Cool, did not know you could do that and it took me a bit to figure out!  Should be good now.

roman-right:
Thank you!

roman-right:
Hi @nzsmith1 

Today is a very special day for me. Beanie 1.0 is out! And this was not possible without your impact.
I'm sending to contributors and people who helped "thank you" letters with stickers.
Could you please share your postal address? I want to send one to you.

My email: roman-right@protonmail.com
Or you can find me on the Beanie discord server: https://discord.gg/ZTTnM7rMaz

PR #32, opened 2021-05-15
Adds unique flag to Indexed field

flyinactor91:
Adds the ability to create or integrate existing unique indexes using the `Indexed` field.

```python
class Plan(Document):
    key: Indexed(str, unique=True)
    name: str
    type: str
```

This is accomplished by turning `_indexed` into a tuple and passing the extra value(s) to `pymongo.IndexModel` as keyword parameters. This can be further extended to add other index attributes to the `Indexed` field init.

This functionality was already available in the code by using a list of `IndexModel` in a model's `Collection` definition, but adding this init kwarg is much more readable and concise. For clarification, it does not affect the index list method demoed in the [index demo](https://github.com/roman-right/beanie-index-demo/blob/28e4db0c1da76abf078cdd9efad9e4a936506205/beanie_index_demo/models/data_models.py#L25).

### DIFF
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -14,14 +14,19 @@ from beanie.odm.operators.find.comparsion import (
 )
 
 
-def Indexed(typ, index_type=ASCENDING):
+def Indexed(typ, index_type=ASCENDING, unique=False):
     """
-    Returns a subclass of `typ` with an extra attribute `_indexed` et to True.
+    Returns a subclass of `typ` with an extra attribute `_indexed` as a tuple:
+    - Index 0 index type set to `pymongo.ASCENDING` or `pymongo.DESCENDING`
+    - Index 1 index unique set to False or True
     When instantiated the type of the result will actually be `typ`.
     """
 
+    if index_type not in (1, -1):
+        raise ValueError(f"Index Type ({index_type}) does not resolve to 1 or -1")
+
     class NewType(typ):
-        _indexed = index_type
+        _indexed = (index_type, unique)
 
         def __new__(cls, *args, **kwargs):
             return typ.__new__(typ, *args, **kwargs)
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -14,19 +14,16 @@ from beanie.odm.operators.find.comparsion import (
 )
 
 
-def Indexed(typ, index_type=ASCENDING, unique=False):
+def Indexed(typ, index_type=ASCENDING, **kwargs):
     """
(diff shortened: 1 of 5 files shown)


flyinactor91:
Additionally, I've added a `ValueError` check for the value of `index_type` since the truthy check in the collection factory now evaluates the tuple's truthiness.

beanie/odm/fields.py
```diff
@@ -14,14 +14,19 @@
...
+def Indexed(typ, index_type=ASCENDING, unique=False):
     """
-    Returns a subclass of `typ` with an extra attribute `_indexed` et to True.
+    Returns a subclass of `typ` with an extra attribute `_indexed` as a tuple:
+    - Index 0 index type set to `pymongo.ASCENDING` or `pymongo.DESCENDING`
+    - Index 1 index unique set to False or True
     When instantiated the type of the result will actually be `typ`.
     """
 
+    if index_type not in (1, -1):
```
roman-right, line 25: There are some other index types, like Text indexes or Geospatial. And other types could be added later. It is better to let the driver check this and not do this on the ODM level.

roman-right (changes requested):
I like your idea. This is a very useful feature.

Could you please add a specific test for this case, which will check if the right index was created with the `unique=True` flag?
I left a comment in the code, please take a look there too.
And it looks like tests are red.

flyinactor91:
To your point about the ODM checking the inputs and looking at the [IndexModel docs](https://pymongo.readthedocs.io/en/stable/api/pymongo/operations.html#pymongo.operations.IndexModel), if the Indexed field uses **kwargs instead of named parameters, it can pass the kwargs directly to the ODM using the same tuple method. Sacrifices some readability but guarantees support for all IndexModel flags even if the ODM adds new ones.

flyinactor91:
In that case, you you prefer a test that only validates that kwargs made it to the IndexModel object or that the Indexed kwargs were successfully applied to the collection's indexes?

roman-right:
Hi @flyinactor91,

Sorry for the delay.

About the index type checking and value error:
I think, this is not needed and it doesn't cover all the cases. Drop this if-statement, please.

About kwargs.
This is an interesting idea. I think this will work, yes. 👍

About tests.
This is up to you. It depends on the implementation, I'd say.

If I will not answer fast enough here, you can reach me in discord.

beanie/odm/fields.py
flyinactor91: Removed check in latest commit

roman-right (approved):

PR #33, opened 2021-05-16
Fix for deprecated import of Mapping ABC

oliverandrich:
> DeprecationWarning: Using or importing the ABCs from 'collections'
> instead of from 'collections.abc' is deprecated since Python 3.3,
> and in 3.10 it will stop working


roman-right:
Hi @oliverandrich ,
Good catch, thank you!

roman-right:
GH actions don't work now. I'll publish a new version after it will be up again. New version PR: https://github.com/roman-right/beanie/pull/34

PR #37, opened 2021-05-17
python 3.10 in actions matrix

roman-right:


roman-right:
Dependencies are not ready for this move yet.

roman-right:
Will try this later

PR #44, opened 2021-06-01
Fixed typo 'comparsion' to 'comparison'

jorants:
Title is self explanatory :-)


roman-right:
Great catch! Thank you!

PR #47, opened 2021-06-11
Added 'save' method and documented save/replace.

jorants:
Although in many cases `update` is the more efficient option, i felt that the undocumented `replace` method might be more natural in many cases. I also added a `save` method which replaces an existing document or inserts a non-existing one and documented both at the top of the update section.


roman-right:
Hi @jorants ,

Sorry for the delay.
Great update! Thank you

Issue #43, opened 2021-06-01
migrations/test_free_fall.py fails


jorants:
Don't know if it is my setup but `migrations/test_free_fall.py` fails on my machine.
Running mongodb v4.4.4 on arch linux, python 3.9.5.

roman-right:
Interesting.
I'll take a look and will add this configuration to the test pipelines.
Thank you!

jorants:
Ah, so it is not generally failing. Here is the error:
```
______________________________ test_migration_free_fall[pyloop] _______________________________

settings = Settings(mongodb_dsn='mongodb://localhost:27017', mongodb_db_name='beanie_db')
notes = None
db = AsyncIOMotorDatabase(Database(MongoClient(host=['localhost:27017'], document_class=dict, tz_aware=False, connect=False, driver=DriverInfo(name='Motor', version='2.4.0', platform='asyncio')), 'beanie_db'))

    async def test_migration_free_fall(settings, notes, db):
        migration_settings = MigrationSettings(
            connection_uri=settings.mongodb_dsn,
            database_name=settings.mongodb_db_name,
            path="tests/migrations/migrations_for_test/free_fall",
        )
>       await run_migrate(migration_settings)

tests/migrations/test_free_fall.py:49: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
beanie/executors/migrate.py:81: in run_migrate
    await root.run(
beanie/migrations/runner.py:73: in run
    await migration_node.run_forward(
beanie/migrations/runner.py:111: in run_forward
    await self.run_migration_class(
beanie/migrations/runner.py:156: in run_migration_class
    await migration.run(session=s)
beanie/migrations/controllers/free_fall.py:26: in run
    await self.function(**function_kwargs)
tests/migrations/migrations_for_test/free_fall/20210413210446_free_fall.py:34: in name_to_title
    await new_note.replace(session=session)
beanie/odm/documents.py:303: in replace
    await self.find_one({"_id": self.id}).replace_one(
beanie/odm/queries/find.py:404: in replace_one
    await self.document_model.get_motor_collection().replace_one(
/usr/lib/python3.9/concurrent/futures/thread.py:52: in run
    result = self.fn(*self.args, **self.kwargs)
[...]

PR #50, opened 2021-06-14
Skip migration test when transactions not avalible

jorants:
Addresses #43 

Not every developer will have mongodb setup using mongos or as a replica set. This skips the failing test when the functionality is not available. I could not test whether it does run the test when it is available as I had to go of documentation, so please check this.

### DIFF
--- a/tests/migrations/test_free_fall.py
+++ b/tests/migrations/test_free_fall.py
@@ -1,6 +1,6 @@
 import pytest
 from pydantic.main import BaseModel
-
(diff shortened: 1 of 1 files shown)


roman-right:
Good point, thank you.
I'll merge this guy in the evening.

roman-right:
Also, just to share how I do this, it is very comfortable to create a free atlas instance to do development things there
https://www.mongodb.com/cloud/atlas
You don't need mongodb (even dockers) on your own system n that case. I like it :-)

PR #54, opened 2021-06-15
Fixed typing erors and added mypy to pre-commit

jorants:
MyPy should now run without errors so I added it to pre-commit.


roman-right:
This looks great! Thank you

PR #55, opened 2021-06-15
test mypy action

roman-right:


jorants:
I think the problem is that the newest version of mypy does detect some errors, i fixed the one in pre-commit to the version in poetry, which seems slightly outdated.

PR #58, opened 2021-06-15
Remove lock file

jorants:
Removes the poetry.lock file.
Due to this mypy can be updated, which caused errors that should be fixed.
I updated mypy in pre-commit as well, and moved the config to pyproject.toml.


roman-right:
It looks like mypy has problems with external modules only here. I'll take a look a little later - have to go of my laptop now.

jorants:
It seems to ignore the pyproject.toml file.
Maybe add `--config-file pyproject.toml` if possible?

roman-right:
> It seems to ignore the pyproject.toml file.
> Maybe add `--config-file pyproject.toml` if possible?

It looks like I can not push to this branch. Could you pls add this parameter into `.github/workflows/github-actions-mypy.yml` ? And remove the version there. The last line should be like:
```
run: mypy beanie/ --config-file pyproject.toml
```

jorants:
Ill try and add the right dependencies to pre-commit as well and push to this branch, wait to merge for a second

jorants:
Type stubs should now work for both precommit and github actions. However, due to [this issue](https://github.com/python-poetry/poetry/issues/3094) with mypy/poetry, they will not be picked up when running mypy through the virtualenv. This should be fixed soon at which point adding `types-toml` and `types-click` to the development dependencies should be enough and we could remove the last block of `pyproject.toml`

roman-right:
> Type stubs should now work for both precommit and github actions. However, due to [this issue](https://github.com/python-poetry/poetry/issues/3094) with mypy/poetry, they will not be picked up when running mypy through the virtualenv. This should be fixed soon at which point adding `types-toml` and `types-click` to the development dependencies should be enough and we could remove the last block of `pyproject.toml`

Great! Thank you for the investigation

PR #72, opened 2021-06-22
Cleanup root dir

jorants:
Removes some files from the root of the repo:

-  Code of conduct is added to documentation
-  Old code of conduct and contribution guidelines are removed.
-  pytest was upgraded to version 6 and now excepts config in pyproject.toml, which was changed to match the script in `scripts`
-  Removed script files. Most where one simple line anyway, only the tests required complicated arguments which are now in the config.


jorants:
I think there was just a networking error with the failed test?

roman-right:
CONTRIBUTING and CODE_OF_CONDUCT files are placed there to follow smth like opensource projects standards. Other services, like snyk.io for example, assume that these files are in the root dir.
 GitHub suggests following these guidelines regarding the topic: https://opensource.guide/

roman-right:
About the tests. Yup, this is a network error. But somehow it happens each time. I'll google, why could it happen. Probably, the matrix in the actions is too big (after I appended pydantic versions there), and something is turning off mongo dockers. Dnt know yet

jorants:
Okay, fair. However, to avoid duplication i re-added them with just a link to the docs, this way they stay the same and up to date.

roman-right:
Nice solution. :+1: 
I'll fix the tests today and merge this.

pyproject.toml
```diff
@@ -54,16 +54,13 @@ exclude = '''
...
 minversion = "6.0"
-addopts = "-ra -q --cov-report term-missing --cov=beanie"
+addopts = "--cov-report term-missing --cov=beanie --cov-branch --cov-fail-under=85"
 testpaths = [
     "tests",
-    "integration",
 ]
 filterwarnings = [
     "error",
-    "ignore::UserWarning",
```
roman-right, line 64: Without this some tests are red.

```>               warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
E               pytest.PytestUnraisableExceptionWarning: Exception ignored in: <socket.socket [closed] fd=-1, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6>
E               
E               Traceback (most recent call last):
E                 File "/home/runner/.cache/pypoetry/virtualenvs/beanie-EwvjBtbN-py3.9/lib/python3.9/site-packages/aiohttp/test_utils.py", line 536, in teardown_test_loop
E                   gc.collect()
E               ResourceWarning: unclosed <socket.socket fd=54, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('127.0.0.1', 60662), raddr=('127.0.0.1', 27017)>
```
Somewhere socket was not closed by the end of the test and it raised a warning. As user warnings are not ignored here, it marked this test as red.

roman-right (changes requested):

roman-right:
Great!

PR #79, opened 2021-07-02
replate pypi homepage from github repo to documentation

KiraPC:
I think that a good think is to have a reference of the Documentation page directly on the PyPI description because mostly user could found the library directly from the package repo and they would directly read the documentation.

The GitHub repository is already present under the property "repository".

Anyway, thanks for this ODM, it is great!


roman-right:
Good point.
On the next version update, this link will be updated.

I send stickers to contributors. Could you please give me your postal address? 
My email is roman-right@protonmail.com
Or you can find me in discord. Beanie server: https://discord.gg/ZTTnM7rMaz

PR #81, opened 2021-07-02
start supporting type annotation for queries results

KiraPC:
Make Query accept a Generic, in order to always know the type of document which they are operating.


roman-right:
Great job!

I've started to clean and test some things a little. You can take a look here: https://github.com/roman-right/beanie/commit/6a19757246272f0c087d31c759f4a94dc9fc5ad5

Will finish and merge ~ on Monday.

KiraPC:
Great!

Just tell me if something is needed.

I guess it could be useful to split implementation from documentation introducing .pyi files.

KiraPC:
I was also thinking:

I had to change the type of *args (the search creteria) of each find method because with the previous type "Mapping[str, Any]" when you send as arguments an expression like "Document.key == 'x'" the hints wont works because does not match the signature, so I changed it to an Union[Mapping[str, Any], Any]. 

But now I guess we can change Any with a custom parameter type called like "ExpressionQuery" that is just a TypeVar that can assume Any type (that's because I don't know how to document the search creteria.)

What do you think? @roman-right

roman-right:
Hi @KiraPC ,

I found a case, which doesn't work for Pylance (vscode default python processor).

You can see the details here: https://github.com/roman-right/beanie/pull/83/checks?check_run_id=2985010908

To reproduce, you can check that branch or just use this code with pyright or pylance:

```python
from typing import List, Dict, Any, Optional

from pydantic import BaseModel

from beanie import Document


class Test(Document):
    foo: str
    bar: str
    baz: str


class ProjectionTest(BaseModel):
    foo: str
    bar: int


async def aggregate_test() -> List[Dict[str, Any]]:
    result = await Test.aggregate([]).to_list()
    result_2 = await Test.find().aggregate([]).to_list()
    return result or result_2


async def aggregate_test_pr() -> List[ProjectionTest]:
    result = (
        await Test.find()
        .aggregate([], projection_model=ProjectionTest)
        .to_list()
    )
    result_2 = await Test.aggregate(
        [], projection_model=ProjectionTest
    ).to_list()
    return result or result_2

```

roman-right:
About the search criteria typing. I think Any is too wide. The thing is, I'm tricking with comparison operators - output is Mapping instead of bool there. MyPy doesn't understand it, and for this subtype of bool could be used, I think. (while original comparison operators output is bool)

roman-right:
Anyway, let's solve this in another branch to not mix issues :-)

KiraPC:
@KiraPC Yes, I added Any because I don't know what I have to use. Anyway, sure, when I have time I'll investigate and work on separate branch. 

Anyway, I'm looking to the issue you mentioned. It is just about aggregate, right?

roman-right:
Yes. Find queries, as I see, work perfectly.

KiraPC:
Ok, perfect.

I have some working stuff to finish now. 

I'll investigate asap.

beanie/odm/documents.py
```diff
@@ -210,7 +259,7 @@ def find_many(
         :param session: Optional[ClientSession] - pymongo session
         :return: [FindMany](https://roman-right.github.io/beanie/api/queries/#findmany) - query instance
         """
-        return FindMany(document_model=cls).find_many(
+        return FindMany[DocType](document_model=cls).find_many(  # type: ignore
```
KiraPC, line 262: @roman-right 

That is one of the two lines I had to disable.

That's because mype throws error like "too few agruments (2 expected)" but that is wrong, it expect just one.

I tried to pass 2 equals arguments, but it fails at runtime

roman-right:
Merged

PR #82, opened 2021-07-02
projection type

roman-right:


roman-right:
implemented in another PR

PR #83, opened 2021-07-04
Kira pc issue 69 fix

roman-right:


roman-right:
Implemented in another PR

PR #89, opened 2021-07-09
Py interface

KiraPC:
I added some pyi files for the modules that has more documentation.

### DIFF
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -8,7 +8,6 @@ from typing import (
     Mapping,
     TypeVar,
     Any,
-    overload,
 )
 
 from bson import ObjectId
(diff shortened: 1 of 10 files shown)


roman-right:
Hey @KiraPC ,
Overall looks good, thank you!

There is a topic to discuss. 
`pyi` files can be placed together with py files or in the separated directory. I didn't decide yet, which way I prefer. 
Do you know something about best practices for this? 
I see pros and cons for both approaches.

Also, it looks like there are some conflicts with the main branch. Pls, pull the last changes from main.

KiraPC:
Hi @roman-right 

Yes, I saw the alternative on how to put the stub files, so I got the decision to put them in the same folder of the source code for this reasons:

- Got inspiration by a famous Python library that is "[numpy](https://github.com/numpy/numpy/tree/main/numpy)", it use the same approach for its interfaces
- For a personal experience. When using a new library, I use to explore the source code directly on the IDE, go to the source of the functionality I'm using by the VSCode feature "go to definition". This feature bring you to the pyi file, if exists, and if the source code is in the same directory, it is easier to have a look on it.

Anyway, lets discuss together what do you prefer.

I'll align with main ASAP.

roman-right:
This guy got outdated. But I keep the idea to make separate interface files alive.

PR #91, opened 2021-07-11
Added bsonable encoder

Vovcharaa:
Added support for all types that could be encoded in bson compatible types + added `bson_encoders` field to Config class for custom encoders.

### DIFF
--- a/beanie/odm/utils/encoder.py
+++ b/beanie/odm/utils/encoder.py
@@ -4,6 +4,7 @@ from enum import Enum
 from pathlib import PurePath
 from types import GeneratorType
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
+from uuid import UUID
 
 from pydantic import BaseModel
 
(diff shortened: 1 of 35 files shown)


Vovcharaa:
For performance reasons where object is bson safe, does this need a switch?
In which section should I put the documentation?

beanie/odm/utils/encoder.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def bsonable_encoder(
+    obj: Any,
+    exclude: Union[
+        AbstractSet[Union[str, int]], Mapping[Union[str, int], Any], None
+    ] = None,
+    by_alias: bool = True,
+    custom_encoder: Dict[Any, Callable[[Any], Any]] = {},
+) -> Any:
```
roman-right, line 50: I think it will return not any, but Mapping of something or Dict of smth. If it is hard to say now, what will be the result, pls add TODO flag here. The same about `obj`

roman-right (changes requested):
Overall looks good. Please take a look at my comments. I think it will be simpler to support the code after some changes.

beanie/odm/utils/encoder.py
```diff
@@ -0,0 +1,122 @@
...
+    ] = defaultdict(tuple)
+    for type_, encoder in type_encoder_map.items():
+        encoders_by_class_tuples[encoder] += (type_,)
+    return encoders_by_class_tuples
+
+
+encoders_by_class_tuples = generate_encoders_by_class_tuples(ENCODERS_BY_TYPE)
+
+
+def bsonable_encoder(
```
roman-right, line 43: Please, add doc strings

roman-right:
Hey @Vovcharaa ,

Sorry for the delay, I'm on vacation.

Thank you for the PR, it looks very good. 
Could you please fix some things? And I'll merge it then

beanie/odm/utils/encoder.py
Vovcharaa: `bsonable_encoder` can accept any type you pass in it and return any type mappend to that type.
It is called recursively on every field except dict keys.

roman-right: I see. Thank you for the clarification

roman-right:
Hey @Vovcharaa ,
Are you ok, if I'll implement the changes I requested myself, or do you want to do it?

Vovcharaa:
> Hey @Vovcharaa ,
> Are you ok, if I'll implement the changes I requested myself, or do you want to do it?

Yes, no problem.
I don't have much time now to do this myself.

roman-right:
Great! Thank you

roman-right:
Hey @Vovcharaa ,
I finally merged this PR :) Thank you for this.
I send Beanie stickers and thx letters to contributors. Could you share your postal address?
My email is roman-right@protonmail.com
Or you can find me in the discord server (info in the Readme here)

Issue #96, opened 2021-07-21
Indexed Field does not support alias


KiraPC:
When creating a new collection with an attribute used as custom index and the value of this attribute is an instance of FieldType, I expect that the index on mongo is created used the Field attribute, instead, it is created with the field name and this could broke the application logic.

```python
import asyncio
from beanie import Document, init_beanie
from motor.motor_asyncio import AsyncIOMotorClient
from pydantic import Field
from beanie.odm.fields import Indexed


class Program(Document):
    class Collection:
        name = "test-program"

    program_id: Indexed(str, unique=True) = Field(..., alias="programId")


async def test():
    new_db_client = AsyncIOMotorClient("mongodb://localhost:27017")

    new_db = new_db_client["testdb"]
    await init_beanie(database=new_db, document_models=[Program])

    program1 = Program(program_id='test-program-1')
    await program1.insert() # OK

    program2 = Program(program_id='test-program-2')
    await program2.insert() # Error: E11000 duplicate key error collection: testdb.test-program index: program_id_1 dup key: { program_id: null }, full error: {'index': 0, 'code': 11000, 'keyPattern': {'program_id': 1}, 'keyValue': {'program_id': None}, 'errmsg': 'E11000 duplicate key error collection: testdb.test-program index: program_id_1 dup key: { program_id: null }'}


asyncio.run(test())
```

The expected result is that both document are stored on the collection.

Instead, removing the alias attribute from Field, it work well, because the document is store with the attribute program_id.

PR #97, opened 2021-07-21
create index by alias if exists

KiraPC:


KiraPC:
Fix for https://github.com/roman-right/beanie/issues/96

beanie/odm/utils/collection.py
```diff
@@ -65,7 +65,13 @@ async def collection_factory(
     # Indexed field wrapped with Indexed()
     found_indexes = [
         IndexModel(
-            [(fname, fvalue.type_._indexed[0])], **fvalue.type_._indexed[1]
+            [
+                (
+                    fvalue.alias if hasattr(fvalue, "alias") else fname,
```
roman-right, line 70: I think, here you can always use `fvalue.alias` . Pydantic already picks the right one there. Look:
https://github.com/samuelcolvin/pydantic/blob/master/pydantic/fields.py#L350

tests/odm/documents/test_init.py
```diff
@@ -59,6 +60,23 @@ async def test_flagged_index_creation():
     }
 
 
+async def test_flagged_index_creation_with_alias():
```
roman-right, line 63: That's great that you made a test for this!

roman-right (commented):
Probably you can avoid extra check when you are choosing a name or alias. Just pick alias each time

beanie/odm/utils/collection.py
KiraPC: Can we assume that fvalue is always a Field instance?

tests/odm/documents/test_init.py
KiraPC: Never commit without tests :D 

(except when I added the typing, never tested type before in python :))

beanie/odm/utils/collection.py
roman-right: I think yes, as we iterate over `__fields__` here which is a dictionary with `ModelField` values only:
https://github.com/samuelcolvin/pydantic/blob/master/pydantic/main.py#L276

KiraPC: Done!

PR #106, opened 2021-08-31
Update annotations for update and delete

tonybaloney:
This solves 3 issues:

1. `await self.find_one({"_id": self.id}).delete(session=session)` would be marked as "delete()" is not awaitable, because the `DeleteOne.__await__()` wasn't annotated as a generator with the correct return type
1. `Documents.find_one()` wasn't annotated causing warnings in other places
1.  Same as 1 but with `update()`.

There is still an issue with the `delete()` methods:

```python
    async def delete(
        self, session: Optional[ClientSession] = None
    ) -> DeleteResult:
        """
        Delete the document

        :param session: Optional[ClientSession] - pymongo session.
        :return: DeleteResult - pymongo DeleteResult instance.
        """
        return await self.find_one({"_id": self.id}).delete(session=session)

    @classmethod
    async def delete_all(
        cls, session: Optional[ClientSession] = None
    ) -> DeleteResult:
        """
        Delete all the documents

        :param session: Optional[ClientSession] - pymongo session.
        :return: DeleteResult - pymongo DeleteResult instance.
        """
        return await cls.find_all().delete(session=session)
```

Technically, `find_all().delete().__await__()` could return `None` since it is a generator. So the return type of `DeleteResult` is not entirely accurate?

### DIFF
--- a/beanie/__init__.py
+++ b/beanie/__init__.py
@@ -4,7 +4,7 @@ from beanie.odm.fields import PydanticObjectId, Indexed
 from beanie.odm.utils.general import init_beanie
 from beanie.odm.documents import Document
 
-__version__ = "1.2.6"
+__version__ = "1.2.7"
 __all__ = [
     # ODM
     "Document",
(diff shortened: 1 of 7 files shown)


roman-right:
Good fix, thank you

> Technically, find_all().delete().__await__() could return None since it is a generator. So the return type of DeleteResult is not entirely accurate?

Hm, looks like yes, you are right. I have to add Optional there

PR #107, opened 2021-08-31
Fix the url about pymongo.IndexModel

grimmer0125:
As title


roman-right:
Good catch! Thank you!

Issue #105, opened 2021-08-30
`Indexed` breaks type checkers


tonybaloney:
The `Indexed` type annotation is raising a warning from Pylance (and other type checkers), because the the result of `Indexed()` is not annotated as any particular type.

```python
class Address(Document):
    street_number: int
    street_name: str
    city: str
    country: str
    longitude: str
    latitude: str
    postal_code: Indexed(str)  # We sort on postal codes, so index them

    class Collection:
        name = "addresses"
```

I _think_, looking at the code, its a `Union[T, _Indexed]` where:

```python
class _Indexed:
    _indexed : Tuple[Any,Dict]

T = TypeVar("T")

def Indexed(typ: T, index_type=ASCENDING, **kwargs) -> Union[T, _Indexed]: 
    ...
```

tonybaloney:
The Pylance error is Illegal type annotation: call expression not allowedPylance

tonybaloney:
@roman-right I have an idea how this could be done, I'll submit a PR if you want

gegnew:
Any thoughts about when this might be fixed?

PR #108, opened 2021-08-31
[WIP] Simplify indexed fields

tonybaloney:
Adds types for indexed fields

```python
class DocumentExample(Document):
   example1 : Indexed(str, pymongo.ASCENDING)
```

would become

```python
class DocumentExample(Document):
   example1 : Indexed(str, pymongo.HASHED)
```

would become

```python
class DocumentExample(Document):
   example1 : HashedIndex[str]
```

* The names of the types are a bit confusing, as they're indexed fields, suggestions?
* Do you like this API?
* This solves #105 
* This could be extended for GeoJSON GeoSphere and Geo2D indexes, but that example is a bit more complicated

### DIFF
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -1,7 +1,10 @@
+from typing import Generic, TypeVar
+
 from bson import ObjectId
 from bson.errors import InvalidId
 from pydantic.json import ENCODERS_BY_TYPE
-from pymongo import ASCENDING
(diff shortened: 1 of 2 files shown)


jsenecal:
Can this PR be fixed to pass the test and get merged at some point ?

roman-right:
Hey @jsenecal 
I forgot about this. Yup, this is a good one. I'll pick it up soon

PR #118, opened 2021-09-23
Feature/on save validation

roman-right:


roman-right:
Docstrings needed

roman-right:
What's inside:

- Before save/save_changes/replace validation
- Revisions support
- Model settings managements

tigrankh (approved):

PR #167, opened 2021-12-14
check if data already Binary

UtkarshMish:
Hi , I have added the necessary code that checks if data , if it is already in binary , then return it in same format.


roman-right:
Hey @UtkarshMish ,
Thank you very much for this PR. I implemented it the same way, but a little in another place here https://github.com/roman-right/beanie/pull/168

PR #178, opened 2021-12-27
Add exclude_hidden to dict()

yallxe:
Previously, when exclude param was provided, hidden fields were not excluded. This commit should fix this issue.

Still, if somebody do not want to exclude hidden fields, he can use `.dict(exclude_hidden=False)`, then, only fields `exclude` will work.

### DIFF
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -1129,6 +1129,7 @@ class Document(BaseModel, UpdateMethods):
         exclude: Union["AbstractSetIntStr", "MappingIntStrAny"] = None,
         by_alias: bool = False,
         skip_defaults: bool = None,
+        exclude_hidden: bool = True,
         exclude_unset: bool = False,
         exclude_defaults: bool = False,
         exclude_none: bool = False,
@@ -1137,8 +1138,8 @@ class Document(BaseModel, UpdateMethods):
         Overriding of the respective method from Pydantic
         Hides fields, marked as "hidden
         """
-        if exclude is None:
-            exclude = self._hidden_fields
+        if exclude_hidden:
+            exclude = self._hidden_fields if exclude is None else {*self._hidden_fields, *exclude}
         return super().dict(
             include=include,
             exclude=exclude,
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -1,6 +1,6 @@
 import asyncio
 import inspect
-from typing import ClassVar
+from typing import ClassVar, AbstractSet
 from typing import (
     Dict,
     Optional,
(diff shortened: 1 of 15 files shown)


roman-right:
Thank you for the PR. It looks very neat.

There is a potential conflict here. Pydantic allows using dicts and sets as excludes:
https://pydantic-docs.helpmanual.io/usage/exporting_models/#advanced-include-and-exclude

If exclude parameter is dict, this merge will use only keys of it. 

Please, add tests for this.

beanie/odm/documents.py
```diff
@@ -1137,8 +1138,8 @@ def dict(
         Overriding of the respective method from Pydantic
         Hides fields, marked as "hidden
         """
-        if exclude is None:
-            exclude = self._hidden_fields
+        if exclude_hidden:
+            exclude = self._hidden_fields if not exclude else {*self._hidden_fields, *exclude}
```
roman-right, line 1142: This merge doesn't cover case, where exclude parameter is dict

roman-right (changes requested):

yallxe:
Well, I've just tried to fix this issue. I think there is a problem with pydantic.

I copied code from https://pydantic-docs.helpmanual.io/usage/exporting_models/#advanced-include-and-exclude to python console (using pydantic 1.8.2) and got an exception:
image

Same issue I have on pydantic 1.7, one that beanie is using.
Python version: 3.10

May I commit my changes that **probably must** allow `exclude` to be dict? Changes include tests also.

roman-right:
You are right, I see the same problem.
I'll make an issue to the pydantic and a PR, if I'll have time this week. Let's wait till this be fixed to not rework twice.

yallxe:
Issue fixed in pydantic 1.9.0
Should beanie migrate to latest pydantic or some tests won't pass?

yallxe:
All tests passed on my hardware, approve me please to run workflows so we can check to check all the versions.

tests/odm/test_fields.py
```diff
@@ -89,9 +103,31 @@ async def test_custom_filed_types():
...
-def test_hidden(document):
-    assert document.revision_id is None
-    assert "revision_id" not in document.dict()
+async def test_hidden(document):
+    document = await DocumentTestModel.find_one()
+
+    assert "test_list" not in document.dict()
+
+
+async def test_param_exclude(document):
```
roman-right, line 112: It is better to use pytest parametrizing here. Using it you can set input values and expected results. With it you will avoid loops and if-else stuff here.
https://docs.pytest.org/en/6.2.x/parametrize.html

tests/odm/test_fields.py
```diff
@@ -1,18 +1,32 @@
...
 
 
+class ExcludeTestCase:
+    def __init__(self, doc: Document, exclude: Union[Mapping, AbstractSet]):
+        self.doc = doc
+        self.exclude = exclude
+
+    def dict(self, *args, **kwargs):
+        x = self.doc.dict(exclude=self.exclude, *args, **kwargs)
+        print(x)
```
roman-right, line 26: kill it :)

roman-right (changes requested):

yallxe:
I've removed `ExcludeTestCase`. Now it looks better for me.

PR #196, opened 2022-02-04
Update allow_index_dropping to be default false

nzsmith1:
- Change default index dropping behaviour to be false
- Add help text to migrations cli
- Do not drop indexes if beanie is not supporting Index creations, the user is probably doing this elsewhere or not at all
- Update docs to add a warning about this parameter so that it is clear to users what occurs when invoked
- Update tests to support changes


beanie/odm/settings/collection.py
```diff
@@ -95,7 +95,8 @@ async def init(
             new_indexes += await collection.create_indexes(found_indexes)
 
         # delete indexes
-        if allow_index_dropping:
+        # Only drop indexes if the user specifically allows for it and Beanie has index fields
+        if allow_index_dropping and found_indexes:
```
roman-right, line 99: It doesn't allow full index dropping for people, who manage MongoDB with Beanie only. I mean the case when somebody decides to not use any indexes for the collection more.

nzsmith1: So my intention was to make this an opt-in process instead of an opt-out process.

There are three cases that I can see:

1. User maintains indexes outside of beanie, no action required and current indexes are maintained
2. User maintains indexes outside of beanie and want to "kick the tires" on beanie creating a new index, the user would add the `index` param to the `Collection` class, no other action required and current indexes are maintained
3. User wants beanie to manage everything and wants beanie to cleanup additional indexes, add `index` param and set `allow_index_dropping` 

In the grand scheme of things having more indexes than you need is not a problem and can easily be taken care of through any of Mongo's tools (compass, atlas, cli, etc), so that is why I wanted to move this to an opt-in on delete.  

Let me know your thoughts on this approach I am happy to change the approach if necessary but my feeling is that deleting indexes should be explicit and opt-in.

roman-right: Sorry for the delay.
Point number 3 doesn't work for the case when there were some indexes in the model before but there are no more. `found_indexes` will be False and indexes will not be deleted even if the user allows it. 

I think the `found_indexes` check is not needed, as the user already allows index dropping explicitly.

PR #197, opened 2022-02-04
feature(state_management): add state_management_replace_objects setting

### LINKED ISSUE #184: Can't replace a dict attribute with save_changes
Hi,

In my project, I need to completely replace a dict attribute, but when using the state management and `save_changes`, the dict are merged instead of replaced. This is due to the logic in the following block: https://github.com/roman-right/beanie/blob/513865aadcf61f57c4ba921ce58bb0e7671036ee/beanie/odm/documents.py#L958-L968

How can I continue to use the state management, yet be able to override a whole dict attribute ?

Thanks for the help,

paul-finary:
This PR fixes #184 by introducing a new setting: `state_management_replace_objects`.


JulienBlancher:
Nice, can't wait for this PR to be merged

PR #201, opened 2022-02-09
Add support for py.typed file

### LINKED ISSUE #194: [mypy] missing library stubs or py.typed marker
Mypy seems to get angry

https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker

nzsmith1:
Resolves #194 

I have added the py.typed file to the project and added it to the pyproject.toml file.  Running a test file:
```
from beanie import Document

class test(Document):
    name: str
    value: int
```

and running the command poetry run mypy test.py, I get the following output with the current build:
`test.py:1: error: Skipping analyzing "beanie": module is installed, but missing library stubs or py.typed marker`

running the same command on this built version, I get the following output:
`Success: no issues found in 1 source file`

So mypy can find the py.typed file just fine now.

I have added a few types where I could, the big section I left as I don't really understand how the field gets converted to a string name is the /odm/operators/*.py field parameters.  This should probably be looked at by someone with a bit more knowledge of the depths than me and can be looked at as a different task.


nzsmith1:
Looks like I did not setup my environment correctly on my home PC, I will have a look at this mypy error tomorrow...

PR #233, opened 2022-03-31
Allow custom MigrationNode for `build`

amos402:
It simple, `MigrationNode.build` use hardcoded `MigrationNode` to create a node object, it forbid its subclasses to reuse this method.


amos402:
Incorrect commit info.

PR #235, opened 2022-03-31
Avoid creating redundant query object

amos402:
Since `motor_cursor` would create a query object every time when it is accessed, this change it's for preventing that meaningless object allocation.


roman-right:
Sounds good.
But it looks like mypy doesn't agree with something. You can run it against your code on committing using pre-commit: `pre-commit install`

amos402:
The first commit was for not to breaking any logic changes.
Yet, the more delightful solution is to set the cursor in `__aiter__`, now it should satisfied the mypy.
Now it can be called by `aiter` multiple times which match the generator standard.

PR #253, opened 2022-05-02
Version: 1.11.0

roman-right:
Features:
- Multi-model
- Views


platipo:
Hi Roman, I'm not sure if this comments belongs here but I got curious and read the PR to understand how you implemented the multi-model.

Basically the idea is adding the `_class_name` field to discriminate against the implementations while querying. Am I correct?

This sparked two ideas:
1. this seem like the general case for the discriminant unions, but while the discriminant unions specify explicitly a string field to specify their type, multi-models use an additional hidden field. This is nice but, thinking of a use case I'm working on, may lead to additional storage consumption, which I couldn't afford.
2. what happens if you change the class name while there are documents in a collection?

platipo:
Hi Roman, I'm not sure if this comments belongs here but I got curious and read the PR to understand how you implemented the multi-model.

Basically the idea is adding the `_class_name` field to discriminate against the implementations while querying. Am I correct?

This sparked two ideas:
1. this seem like the general case for the discriminant unions, but while the discriminant unions specify explicitly a string field to specify their type, multi-models use an additional hidden field. This is nice but, thinking of a use case I'm working on, may lead to additional storage consumption, which I couldn't afford.
2. what happens if you change the class name while there are documents in a collection?

roman-right:
Hey @platipo 

1. Unfortunately this is the only way I came up with to be sure, that I can identify the class correctly
2. In one of the next updates I'll add a config for this to specify the mapping name, it will be possible to change the class name then

Issue #133, opened 2021-10-19
Distinct method?


tgrandjean:
Hello,

What a wonderfull work you've done! I'm starting using Beanie with fastapi and I couldn't find anything in the docs or in the code that allow me to use the `db.collection.disctinct` from [mongodb docs](https://docs.mongodb.com/manual/reference/method/db.collection.distinct/).

I'm using the following document:

```python
class Customer(BaseModel):
    name: str
    email: Optional[str]
    # some fields

class Invoice(Document)

   customer: Customer
    # some fields
```

Coming from pymongo, I use `collection.distinct('customer')` to get the list of distinct customers from the collection 'invoices', is there a way to do this (or something equivalent using aggregate and project) with Beanie?

nichitag:
Roman, how about this future?

nichitag:
https://github.com/roman-right/beanie/pull/260

PR #268, opened 2022-05-12
implement distinct future #133

gruianichita:


gruianichita:
https://github.com/roman-right/beanie/issues/133

PR #273, opened 2022-05-18
New feature - Check for unique fields

Mohsen-Khodabakhshi:
Hi. I needed to check the duplicate fields(unique fields) in the database. I added unique_fields to ItemSettings to solve this.
Sample:
```
class User(Document):
    first_name: str | None = None
    last_name: str | None = None
    username: str
    national_code: str

    class Settings:
        name = "users"
        unique_fields = ['username', 'national_code'] # list of your model unique fields
```
If each is a duplicate in the database, It will raise UniqueFieldExists exception.

### DIFF
--- a/beanie/odm/models.py
+++ b/beanie/odm/models.py
@@ -2,8 +2,14 @@ from typing import List
 
 from pydantic import BaseModel
 
+from beanie.exceptions import UniqueFieldExists
 from beanie.odm.enums import InspectionStatuses
 from beanie.odm.fields import PydanticObjectId
+from beanie.odm.actions import (
+    Insert,
+    Replace,
+    before_event,
+)
 
 
 class InspectionError(BaseModel):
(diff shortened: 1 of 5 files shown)


beanie/odm/models.py
```diff
@@ -22,3 +28,17 @@ class InspectionResult(BaseModel):
 
     status: InspectionStatuses = InspectionStatuses.OK
     errors: List[InspectionError] = []
+
+
+class UniqueFieldsModel: # noqa
+    @before_event([Insert, Replace])
+    async def check_for_unique_fields(self):
+        unique_fields = self.get_unique_fields()
+        if unique_fields:
```
cikay, line 37: Reverse the condition and return early to decrease the indentation level.

cikay:
@roman-right Why do you not take care of this feature?

roman-right:
Hi!
Sorry for the delay.
Could you pls tell me if don't unique indexes cover this feature? Thank you

PR #275, opened 2022-05-29
Turn the caching system into an interface to support different caching strategies

efirvida:
I thought it would be a good idea to have several caching strategies, so I modified a bit the cache API to give ideas of how it could be, for example, to use Redis as a caching system.

```python
import redis
from beanie import Document
from beanie.odm.cache import RedisCache, LRUCache

redis_client = redis.Redis(host='localhost', port=6379, db=0)

class Product(Document):
    name: str
    ...

    class Settings:
        use_cache = True
        cache_system = RedisCache(redis_client)

class User(Document):
    name: str
    ...

    class Settings:
        use_cache = True
        cache_system = LRUCache(
            capacity=32, expiration_time=timedelta(seconds=10)
        )

```

### DIFF
--- a/beanie/odm/cache.py
+++ b/beanie/odm/cache.py
@@ -127,8 +127,6 @@ class RedisCache(Cache):
         return data if data else None
 
     def set(self, key, value) -> None:
-        if not value:
-            return
         value = self._serialize(value)
         key = self._get_key(key)
         self.redis_client.set(key, value, ex=self.ttl)
--- a/beanie/odm/cache.py
+++ b/beanie/odm/cache.py
@@ -25,6 +25,10 @@ class Cache(abc.ABC):
     def get(self, key: str) -> Optional[CachedItem]:
         ...
 
+    @abc.abstractmethod
+    def invalidate(self) -> None:
+        ...
+
     @staticmethod
     def create_key(*args):
         return str(args)  # TODO think about this
@@ -56,6 +60,9 @@ class LRUCache(Cache):
             if len(self.cache) >= self.capacity:
                 self.cache.popitem(last=False)
         self.cache[key] = CachedItem(value=value)
+
+    def invalidate(self) -> None:
+        self.cache.clear()
 
 
 class RedisCache(Cache):
@@ -125,3 +132,6 @@ class RedisCache(Cache):
         value = self._serialize(value)
         key = self._get_key(key)
         self.redis_client.set(key, value, ex=self.ttl)
+
+    def invalidate(self) -> None:
+        self.redis_client.flushdb()
--- a/beanie/odm/cache.py
+++ b/beanie/odm/cache.py
@@ -1,3 +1,4 @@
+import abc
 import collections
 import datetime
 from datetime import timedelta
@@ -13,7 +14,21 @@ class CachedItem(BaseModel):
     value: Any
 
 
-class LRUCache:
+class Cache(abc.ABC):
+    @abc.abstractmethod
+    def set(self, key: str, value: Any) -> None:
+        ...
+
+    @abc.abstractmethod
+    def get(self, key: str) -> Optional[CachedItem]:
+        ...
+
+    @staticmethod
+    def create_key(*args):
+        return str(args)  # TODO think about this
+
+
+class LRUCache(Cache):
     def __init__(self, capacity: int, expiration_time: timedelta):
         self.capacity: int = capacity
         self.expiration_time: timedelta = expiration_time
@@ -40,6 +55,3 @@ class LRUCache:
                 self.cache.popitem(last=False)
         self.cache[key] = CachedItem(value=value)
 
-    @staticmethod
-    def create_key(*args):
-        return str(args)  # TODO think about this
--- a/beanie/odm/cache.py
+++ b/beanie/odm/cache.py
@@ -1,9 +1,11 @@
 import abc
 import collections
 import datetime
+import pickle
 from datetime import timedelta
-from typing import Any, Optional
+from typing import Any, Optional, Sequence, Union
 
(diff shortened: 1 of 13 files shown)


roman-right:
Oh wow! Great! I'll check everything and merge it tomorrow. Thank you for this feature

roman-right:
It looks like it breaks some tests

efirvida:
> It looks like it breaks some tests

Hi!, I'm, not surprised. This was just an idea to work on, that came to me on a project that I want to use Beanie. And also because  I'm learning about Redis as a caching system.

I did not expect such a quick and positive reaction to my proposal. I'm going to work it better to make it ready for merge. Because even the RedisCashe implementation is very poor at this point.

efirvida:
@roman-right, I have made some fixes and improved de Redis cache system with msgpack as serialization and add ttl.

But there is something strange that my first commit brake (f6614c4ccf4ed8613c70c1d23d174c4af6ef3f79) some test "no cache-related" from my point of view its the same thing but not for the test.

if I change this:

```
    @classmethod
    def init_cache(cls) -> None:
        """
        Init model's cache
        :return: None
        """
        if cls.get_settings().use_cache:
            cls._cache = cls.get_settings().cache_system

```
 for this:

```
    @classmethod
    def init_cache(cls) -> None:
        """
        Init model's cache
        :return: None
        """
        if cls.get_settings().use_cache:
            cls._cache = LRUCache(capacity=5, expiration_time=datetime.timedelta(seconds=10))

```

Which I think is the same thing all test runs, but not if I init the cache from the Document's `cache_system` settings

efirvida:
> 0c9d28e

Fixed here [0c9d28e](https://github.com/roman-right/beanie/pull/275/commits/0c9d28e00a54731df0774e1ba3f611b03356485f)

efirvida:
Now it's working, here is an example code

```python

import redis
import msgpack
import json
import pickle
from typing import Optional
from datetime import timedelta

from beanie.odm.cache import LRUCache,RedisCache

redis_client = redis.Redis(host='localhost', port=6379, db=0)

class ProductNoCache(Document):
    name: str
    description: Optional[str] = None
    price: Indexed( float ) 
    class Settings:
        use_cache = False


class ProductRedisPickle(Product):
    class Settings:
        use_cache = True
        cache_system = RedisCache(redis_client, serializer=pickle) # pickle is default serializer


class ProductRedisJSON(Product):
    class Settings:
        use_cache = True
        cache_system = RedisCache(redis_client, serializer=json, ttl=timedelta(seconds=10)) # use ttl for expiration
        
       
class ProductRedisMsgPack(Product):
    class Settings:
        use_cache = True
        cache_system = RedisCache(redis_client, serializer=msgpack, ttl=10) # ttl can be int -> 10 seconds


class ProductLRUCache(Product):
    class Settings:
        use_cache = True
        cache_system = LRUCache(capacity=5, expiration_time=timedelta(seconds=10))```

**Comment by roman-right** [comment]
Wow! You worked a lot here. Thank you!
I'll check everything this weekend.

File: beanie/odm/cache.py [hunk 1]
```diff
@@ -40,6 +61,75 @@ def set(self, key, value) -> None:
...
+        serializer=pickle,
+    ):
+        self.redis_client = redis_client
+        self.key_prefix = key_prefix
+        self.ttl = self._normalize_ttl(ttl)
+        self.serializer = serializer
+
+    def _normalize_object_ids(self, value):
+        if isinstance(value, dict):
+            if "_id" in value:
```
roman-right, line 83: I would use a type-based custom serializer for msgpack or json

pyproject.toml
```diff
@@ -20,6 +20,8 @@ motor = "^2.5"
 click = ">=7"
 toml = "*"
 yarl = ">=1.6"
+redis = "^4.3.1"
```
roman-right, line 23: it would be great to have these guys as extra dependencies: https://python-poetry.org/docs/pyproject#extras

tests/odm/documents/test_bulk_write.py
```diff
@@ -22,6 +22,7 @@ async def test_insert(documents_not_inserted):
 
 async def test_update(documents, document_not_inserted):
     await documents(5)
+    DocumentTestModel.invalidate_cache()
```
roman-right, line 25: as this is not a part of the test, it is better to move it as a fixture to the modular conftest file and run it in `autouse=True` mode

beanie/odm/cache.py
```diff
@@ -40,6 +61,75 @@ def set(self, key, value) -> None:
...
+            loaded = self.serializer.loads(value)
+            return loaded
+        except (
+            TypeError,
+            pickle.UnpicklingError,
+            UnicodeDecodeError,
+            msgpack.exceptions.UnpackException,
+            msgpack.exceptions.ExtraData,
+        ):
+            return None
```
roman-right, line 110: I didn't get the logic here. I think, users should know, that the cache is broken. May be I don't see something

roman-right (changes requested):

beanie/odm/cache.py
```diff
@@ -40,6 +61,75 @@ def set(self, key, value) -> None:
...
+        redis_data = self.redis_client.get(key)
+        data = self._deserialize(redis_data)
+        return data if data else None
+
+    def set(self, key, value) -> None:
+        value = self._serialize(value)
+        key = self._get_key(key)
+        self.redis_client.set(key, value, ex=self.ttl)
+
+    def invalidate(self) -> None:
```
Luc1412, line 134: All Redis action should be async as beanie is async too. This could introduce blocking behaviour.

Luc1412 (changes requested):

PR #291, opened 2022-06-21
Adding Update Action

schwannden:
When using `Document.update` or `Document.set` (which triggers Document.update), it will trigger `Update` action and its registered hook. This has the benefit of allowing people who use update or set to also use actions to implement hooks like updating `updated_at` timestamp.

### DIFF
--- a/.github/workflows/github-actions-publish-docs.yml
+++ b/.github/workflows/github-actions-publish-docs.yml
@@ -12,9 +12,9 @@ jobs:
         with:
           python-version: 3.9
       - name: install poetry
-        uses: abatilo/actions-poetry@v2.0.0
+        uses: abatilo/actions-poetry@v2
         with:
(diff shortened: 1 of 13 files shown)


schwannden:
@roman-right thank you for this great package, this is a small pr that updated both code, doc, and tests. Should not take too long to review, but will help our use case a lot, thank you.

schwannden:
@roman-right any update?

schwannden:
The reason why we need this, is because in our API design, we have a schema for POST body json, and a separate schema for PUT body json. The one for PUT is basically changing all the fields in POST to optional. This means we want to allow partial update, therefore easier if we use set or update function instead of save or insert. But in this case, we still want to makesure our updated_at field is updated.

roman-right:
Hey! Sorry for the delay. I plan to merge this tomorrow

PR #299, opened 2022-07-03
fix: add count support for aggregations

harryrobertwright:
Calling `.count()` on a `FindMany` object that was constructed with the `fetch_links` parameter does not work properly, as the function does not currently make use of motor's `.aggregate(...)`. This PR aims to fix that by appending the aggregation pipeline and drawing the count from there.

Hopefully this is helpful - love the project!

### DIFF
--- a/tests/odm/models.py
+++ b/tests/odm/models.py
@@ -67,6 +67,7 @@ class DocumentTestModel(Document):
     test_list: List[SubDocument] = Field(hidden=True)
     test_doc: SubDocument
     test_str: str
+    test_link: Optional[Link[Document]] = None
 
     class Settings:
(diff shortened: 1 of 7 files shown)


tests/odm/models.py
```diff
@@ -67,6 +67,7 @@ class DocumentTestModel(Document):
     test_list: List[SubDocument] = Field(hidden=True)
     test_doc: SubDocument
     test_str: str
+    test_link: Optional[Link[Document]] = None
```
roman-right, line 70: Please, use a separate model for testing this case

harryrobertwright:
Think this issue has been fixed in subsequent releases so will close.

Issue #225, opened 2022-03-23
Implement `Delete` action event handler


mdgilene:
Events exist for the basic Insert/Update events. Main event missing is the `Delete` event. This would be useful for executing a variety of cleanup actions after an instance is deleted.

roman-right:
Hey!
Hm, I do agree. I'll check if it will not conflict with any other code and will add this soon.

roman-right:
It is available in the `1.10.3`. Docs will be updated soon too

PR #303, opened 2022-07-12
doc(events): fix formatting and add documentation for Delete event

paul-finary:
Hi,

I saw that the formatting was wrong in the doc, so I fixed it, and added the documentation for the delete event added following [this](https://github.com/roman-right/beanie/issues/225) issue.

Best,


paul-finary:
Hi @roman-right,

I just fine-tuned the documentation for events.

Best,

roman-right:
Nice! Thank you! :)

PR #343, opened 2022-09-06
Timestamp field of MongoDB 

miladvayani:
Timestamp field added to fields and we can save timestamp in db with timestamp type.


roman-right:
Hi! Thank you for the PR. Could you pls provide some tests for this improvement?

Issue #300, opened 2022-07-03
Feature request: Support for Optional[List[Link]] relation


SamEdwardes:
## Feature request

Thank you for all of the great work on Beanie @roman-right! I would like to make a feature request. It would be helpful to extend the supported relationship patterns to include:

```
Optional[List[Link[...]]
```

Per the [docs](https://roman-right.github.io/beanie/tutorial/relations/) the currently supported patterns are:

- `Link[...]`
- `Optional[Link[...]]`
- `List[Link[...]]`

## Related issues

- #220

rgajason:
+1

This feature would be helpful for situations where the parent object is created separately from the process to add/remove items to it.  For example, a "group" model that links to a list of "users".  It is valid for a group to exist and be empty.

roman-right:
Hi!
Thank you for the suggestion.
I'll add this soon.

turkic-dev:
@roman-right When will this feature be added?

I'm using it like this right now, would that be a problem? `perm:List[Link[Permission]]  = []`

SamEdwardes:
> @roman-right When will this feature be added?
> 
> I'm using it like this right now, would that be a problem? `perm:List[Link[Permission]] = []`

I think you will want to use the `Field` class like this:

```python
from pydantic import Field

perm:List[Link[Permission]] = Field(default_factory=list)
```

PR #358, opened 2022-09-27
Feature/optional list link

rga-alex-deng:
Attempt to implement feature for an additional relation field

Optional[List[Link[]]]

See issue: https://github.com/roman-right/beanie/issues/300

I was hoping I could get some feedback on this feature if possible @roman-right 

I'm aware that I'm missing some integration testing in the code however we integrated and tested it on a different application and we're not seeing any issues so far.

Thanks for your time!

### DIFF
--- a/beanie/__init__.py
+++ b/beanie/__init__.py
@@ -26,7 +26,7 @@ from beanie.odm.documents import Document
 from beanie.odm.views import View
 from beanie.odm.union_doc import UnionDoc
 
-__version__ = "1.11.12"
+__version__ = "1.12.0"
 __all__ = [
     # ODM
     "Document",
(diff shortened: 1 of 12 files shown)


roman-right:
Hey!

Thank you for the PR. The code looks good!

Could you please:
- fix the linter issues. You can check it using pre-commit.
- update the documentation. Here is the place to add the new type: https://github.com/roman-right/beanie/blob/main/docs/tutorial/relations.md

Thank you!

rga-alex-deng:
> Hey!
> 
> Thank you for the PR. The code looks good!
> 
> Could you please:
> 
> * fix the linter issues. You can check it using pre-commit.
> * update the documentation. Here is the place to add the new type: https://github.com/roman-right/beanie/blob/main/docs/tutorial/relations.md
> 
> Thank you!

Thank you for the feedback @roman-right !

I've fixed the linter issues and updated the doc. Changes have been pushed to the branch on this PR.

PR #417, opened 2022-11-14
chore: Prevent deprecation warning when `skip_defaults` is not explicitly set

jtfidje:
`skip_defaults` has been deprecated in Pydantic and is causing a lot of warnings in my tests.
This PR will remove these warnings if `skip_defaults` is not explicitly set, but won't break the API if someone still use the argument.

### DIFF
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -874,15 +874,21 @@ class Document(
                 )  # type: ignore
             elif exclude is None:
                 exclude = self._hidden_fields
-        return super().dict(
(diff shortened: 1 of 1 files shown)


flinz:
TFTPR! I was just about to start a PR to fix this, it's popping up a lot in tests.

jtfidje:
My pleasure 😃 

Sounds like a good idea!

roman-right:
Thank you for the PR. Merging. It will be available in the next version.

PR #428, opened 2022-11-19
Implemented event type propagation to listeners

Vitalium:
This feature allows to declare functions that are event listeners with additional `event` parameter.
Example:

```python
class Sample(Document):
    num: int
    name: str
    last_event: str = ''

    @before_event(Insert, Replace, Update)
    def log_event(self, event: EventTypes):
        if event is Insert:
            print('Sample document created!')
        elif event in (Replace, Update):
            print('Sample document changed!')

        self.last_event = str(event.value)
```

This implementation is backward compatible, any functions will be called properly based on their signatures.

_Inspired by DaveRGA's question on Beanie Discord server at Nov, 17 2022._

### DIFF
--- a/beanie/odm/actions.py
+++ b/beanie/odm/actions.py
@@ -115,6 +115,7 @@ class ActionRegistry:
         :param instance: Document - object of the Document subclass
         :param event_type: EventTypes - event types
         :param action_direction: ActionDirections - before or after
+        :param exclude: list[ActionDirections | str] - list of ActionDirections or action names to skip
         """
         if action_direction in exclude:
             return
@@ -128,10 +129,13 @@ class ActionRegistry:
             if action.__name__ in exclude:
                 continue
 
+            args = (instance, event_type)
+            count = action.__code__.co_argcount  # type: ignore
+
             if inspect.iscoroutinefunction(action):
-                coros.append(action(instance))
+                coros.append(action(*args[:count]))
             elif inspect.isfunction(action):
-                action(instance)
+                action(*args[:count])
 
         await asyncio.gather(*coros)
 
--- a/beanie/odm/utils/init.py
+++ b/beanie/odm/utils/init.py
@@ -21,6 +21,29 @@ from beanie.odm.union_doc import UnionDoc
 from beanie.odm.utils.relations import detect_link
 from beanie.odm.views import View
 
+ALL_TYPES_EXCEPT_NULL = [
+    -1,
+    1,
+    2,
+    3,
+    4,
+    5,
+    6,
+    7,
+    8,
+    9,
+    11,
+    12,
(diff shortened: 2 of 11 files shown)


Vitalium:
Hi! It seems I make something wrong as the second commit is also included in this PR.

beanie/odm/utils/init.py
```diff
@@ -258,19 +281,31 @@ async def init_indexes(cls, allow_index_dropping: bool = False):
...
-                    (
-                        fvalue.alias,
-                        fvalue.type_._indexed[0],
+        found_indexes = []
+        for fvalue in cls.__fields__.values():
+            if getattr(fvalue.type_, "_indexed", False):
+                kwargs = fvalue.type_._indexed[1]
+
+                # if indexed field is Optional, index creation should be filtered
+                print(
```
roman-right, line 290: Please, remove this

beanie/odm/utils/init.py
```diff
@@ -21,6 +21,29 @@
 from beanie.odm.utils.relations import detect_link
 from beanie.odm.views import View
 
+ALL_TYPES_EXCEPT_NULL = [
```
roman-right, line 24: This needs an explanation in the inline comment with a reference to the doc to make it supportable. Or, if there is such a thing but with 0 in the pymongo package, it is better to construct this one based on it.

beanie/odm/actions.py
```diff
@@ -128,10 +129,13 @@ async def run_actions(
             if action.__name__ in exclude:
                 continue
 
+            args = (instance, event_type)
+            count = action.__code__.co_argcount  # type: ignore
```
roman-right, line 133: This looks hacky. Let's think if it is possible to avoid it. If I'll have ideas, I'll put it here

roman-right (changes requested):
Pls look at the comments

beanie/odm/utils/init.py
Vitalium: Yes, it was accidentally appeared here, this is another feature implementation.

beanie/odm/utils/init.py
Vitalium: This is a list of standard BSON type IDs: https://www.mongodb.com/docs/manual/reference/bson-types/ I'll add a link there.

beanie/odm/actions.py
Vitalium: Python at all is a magical hacky language :)
Another option is to use `__annotation__` but it can be unset if no type hints used.

Vitalium: I suppose there is a way to do it like this: `count = len(inspect.signature(action)._parameters)`

PR #466, opened 2022-12-29
Allow change class_id and use name settings in UnionDoc

wonjoonSeol-WS:
Beanie is all sound and good until there is a need to connect to existing DB where beanie is/was not used.
For instance, one might have SNS -> SQS -> lambda (that uses any other client other than beanie) -> DocDB / Mongo. 
In case like this, the existing data would not have beanie's internal data field such as '_class_id".

This makes switching to Beanie difficult as it requires full DB migration, especially if DB is used for event-sourcing. 

This PR allows (1) custom internal data field name to be set and (2) custom name to be set in Union doc children classes.

Usage:
```
class Parent(UnionDoc):
    class Settings:
        class_id = "event_type" <-
        name = "eventsource"

class ChildCreated(Document):
    event_type: Literal["created"] = "created" <- event_type is used instead of _class_id 
    sent_at: datetime

    class Settings:
        name = "created" <- This was option was ignored before
        union_doc = Parent

class ChildDeleted(Document):
    event_type: Literal["created"] = "deleted" <- event_type is used instead of _class_id 
    sent_at: datetime

    class Settings:
        name = "deleted" <- This was option was ignored before
        union_doc = Parent
```

Example:

Schema
```
class Parent(UnionDoc):
    class Settings:
        class_id = "event_type"
        name = "collection"


class One(Document):
    test: int

    class Settings:
        name = "two"
        class_id = "event_type"
        union_doc = Parent
        
 class Two(Document): <- without union_doc
    test: int

    class Settings:
        name = "three"
        class_id = "event_type"

```

Test
```
In [1]: await db.One(test=1).insert()
Out[1]: One(id=ObjectId('63ae1890edab780d45df0b07'), revision_id=None, test=1)

pymongo
In [17]: list(c["db"]["collection"].find({'test': 1}))
Out[17]: [{'_id': ObjectId('63ae1890edab780d45df0b07'), 'event_type': 'two', 'test': 1}]

In [1]: await db.Two(test=3).insert()
Out[1]: Two(id=ObjectId('63ae19cd8e684a4cd769fb1f'), revision_id=None, test=3)

pymongo
In [23]: list(c["db"]["three"].find({'test': 3}))
Out[23]: [{'_id': ObjectId('63ae19cd8e684a4cd769fb1f'), 'test': 3}]

```

---

FYI, https://github.com/roman-right/beanie/pull/206 this PR was not reviewed for a year so I closed it..

### DIFF
--- a/beanie/odm/interfaces/find.py
+++ b/beanie/odm/interfaces/find.py
@@ -427,5 +427,5 @@ class FindInterface:
                 )
 
         if cls.get_settings().union_doc:
-            args += ({"_class_id": cls.__name__},)
+            args += ({"_class_id": cls.get_settings().class_id},)
         return args
--- a/beanie/odm/interfaces/find.py
+++ b/beanie/odm/interfaces/find.py
@@ -415,11 +415,11 @@ class FindInterface:
             and cls._inheritance_inited
         ):
             if not with_children:
-                args += ({"_class_id": cls._class_id},)
+                args += ({cls.get_settings().class_id: cls._class_id},)
             else:
                 args += (
                     {
-                        "_class_id": {
+                        cls.get_settings().class_id: {
                             "$in": [cls._class_id]
                             + [cname for cname in cls._children.keys()]
                         }
--- a/beanie/odm/interfaces/find.py
+++ b/beanie/odm/interfaces/find.py
@@ -405,7 +405,7 @@ class FindInterface:
             (
                 True
                 for a in args
-                if isinstance(a, Iterable) and "_class_id" in a
+                if isinstance(a, Iterable) and cls.get_settings().class_id in a
             )
         ):
             return args
--- a/beanie/odm/interfaces/find.py
+++ b/beanie/odm/interfaces/find.py
@@ -427,5 +427,5 @@ class FindInterface:
                 )
 
         if cls.get_settings().union_doc:
-            args += ({"_class_id": cls.get_settings().class_id},)
+            args += ({"cls.get_settings().class_id": cls.get_settings().name or cls.__name__},)
         return args
--- a/beanie/odm/interfaces/find.py
+++ b/beanie/odm/interfaces/find.py
@@ -427,5 +427,5 @@ class FindInterface:
                 )
 
         if cls.get_settings().union_doc:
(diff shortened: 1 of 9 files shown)


beanie/odm/utils/init.py
```diff
@@ -233,8 +233,9 @@ async def init_document_collection(self, cls):
         # register in the Union Doc
 
         if document_settings.union_doc is not None:
+            name = cls.get_settings().name or cls.__name__
```
wonjoonSeol-WS, line 236: if `class.__name__` returns original class name instead of None when get_settings().name is not set, we wouldn't need this line.

beanie/odm/settings/base.py
```diff
@@ -18,6 +18,7 @@ class ItemSettings(BaseModel):
     motor_collection: Optional[AsyncIOMotorCollection] = None
 
     union_doc: Optional[Type] = None
+    class_id: str = "_class_id"
```
wonjoonSeol-WS, line 21: For now, this setting would be ignored in Document class type

beanie/odm/utils/parsing.py
```diff
@@ -24,7 +24,7 @@ def parse_obj(
             raise UnionHasNoRegisteredDocs
 
         if isinstance(data, dict):
-            class_name = data["_class_id"]
+            class_name = data[model.get_settings().class_id]
         else:
             class_name = data._class_id
```
wonjoonSeol-WS, line 29: Does it ever reach this line? it didn't for my (small) dataset

beanie/odm/interfaces/find.py
```diff
@@ -427,5 +427,5 @@ def _add_class_id_filter(cls, args: Tuple, with_children: bool = False):
                 )
 
         if cls.get_settings().union_doc:
-            args += ({"_class_id": cls.__name__},)
+            args += ({"cls.get_settings().class_id": cls.get_settings().name or cls.__name__},)
```
wonjoonSeol-WS, line 430: Fixing in progress

roman-right:
Hi @wonjoonSeol-WS ,
Sounds very reasonable. I'll check the PR soon this week

roman-right:
Merged. Thank you very mush for the PR.
It will be published in the next release (beta or normal) today-tomorrow

PR #478, opened 2023-01-28
fix: make `revision_id` not show in schema

### LINKED ISSUE #418: [BUG] revision_id is in schema even it should be removed by Document.schema_extra
**Describe the bug**
revision_id still in the schema and since it's hidden it should not appear in the schema
`hidden` is not compliant with OpenAPI/JSONSchema

**To Reproduce**
```python
class MyModel(Document):
    a:int

print(MyModel.get_hidden_fields())
print(MyModel.schema_json(indent=2))
```
```python
{'revision_id'}
```
```json
{
  "title": "MyModel",
  "description": "Document Mapping class.\n\nFields:\n\n- `id` - MongoDB document ObjectID \"_id\" field.\nMapped to the PydanticObjectId class\n\nInherited from:\n\n- Pydantic BaseModel\n- [UpdateMethods](https://roman-right.github.io/beanie/api/interfaces/#aggregatemethods)",
  "type": "object",
  "properties": {
    "_id": {
      "title": " Id",
      "type": "string",
      "examples": [
        "5eb7cf5a86d9755df3a6c593",
        "5eb7cfb05e32e07750a1756a"
      ]
    },
    "revision_id": {
      "title": "Revision Id",
      "hidden": true,
      "type": "string",
      "format": "uuid"
    },
    "a": {
      "title": "A",
      "type": "integer"
    }
  },
  "required": [
    "a"
  ]
}
```

**Expected behavior**
since the field is hidden it should not appear in the schema
this causes some OpenAPI tools to crash, e.g: generating a client
```shell
docker run  --network host  --rm -v "{PWD}:/local" openapitools/openapi-generator-cli generate \
    -i local/openapi.json \
		-g python \
		-o /local/out/python

Exception in thread "main" org.openapitools.codegen.SpecValidationException: There were issues with the specification. The option can be disabled via validateSpec (Maven/Gradle) or --skip-validate-spec (CLI).
 | Error count: 3, Warning count: 0
Errors: 
	-attribute components.schemas.Model.hidden is unexpected
	-attribute paths.'/Model'(get).parameters.[revision_id].schemas.hidden is unexpected
	-attribute components.schemas.Model.hidden is unexpected
```

ivan-gj:
Fix #418

This is my first ever pull request (will hopefully not be the last), so please be kind!  :sweat_smile: 

I basically found [this solution](https://github.com/tiangolo/fastapi/issues/1378#issuecomment-1374968471) and applied it to the `Document` class, and it worked at first try. So, thanks @Xenthys!

Also, @roman-right, correct me if I am wrong but I believe the previous two lines: 
```python
            for field_name in model._hidden_fields:
                schema.get("properties", {}).pop(field_name, None)
```
become redundant. I believe they were trying to accomplish the same thing, but it seems that it only works once the doc is initialized.

So, if you agree @roman-right, I can remove the redundancy before merging.

### DIFF
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -997,3 +997,9 @@ class Document(
         ) -> None:
             for field_name in model._hidden_fields:
                 schema.get("properties", {}).pop(field_name, None)
+
+            props = {}
+            for k, v in schema.get('properties', {}).items():
+                if not v.get("hidden", False):
+                    props[k] = v
+            schema["properties"] = props
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -995,11 +995,8 @@ class Document(
         def schema_extra(
             schema: Dict[str, Any], model: Type["Document"]
         ) -> None:
-            for field_name in model._hidden_fields:
-                schema.get("properties", {}).pop(field_name, None)
-
             props = {}
-            for k, v in schema.get('properties', {}).items():
+            for k, v in schema.get("properties", {}).items():
                 if not v.get("hidden", False):
                     props[k] = v
             schema["properties"] = props
--- a/tests/odm/test_fields.py
+++ b/tests/odm/test_fields.py
@@ -5,6 +5,8 @@ from typing import Mapping, AbstractSet
 import pytest
 from pydantic import BaseModel, ValidationError
 
+from beanie import Document
+from beanie.exceptions import CollectionWasNotInitialized
 from beanie.odm.fields import PydanticObjectId
 from beanie.odm.utils.dump import get_dict
 from beanie.odm.utils.encoder import Encoder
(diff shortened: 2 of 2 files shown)


roman-right:
Hey! Thank you for the PR. I'll take a look tomorrow

NickRafRGA:
I'm running into this issue as well, any timeline on when this will be merged in?

roman-right:
Hi guys,
Sorry for the delay.
Could you please provide an example of when the `revision_id` appears in the schema? I need it to add a test for this case.
@ivan-gj , @NickRafRGA

ivan-gj:
No worries, @roman-right, here's a small test written for `pytest`: 
```python3
from beanie import Document


class Foo(Document):
    """Dummy document."""
    bar: int = 3


def test_revision_id_not_in_schema():
    """Check if there is a `revision_id` slipping into the schema."""
    schema = Foo.schema()
    assert "revision_id" not in schema["properties"]
```

ivan-gj:
Actually, here's a version with an extra check (just in case) where we actually specify that the test is only valid if the collection is not initialized (if the collection is initialized then the issue is gone):
```python3
from beanie import Document
from beanie.exceptions import CollecctionWasNotInitialized

import pytest


class Foo(Document):
    """Dummy document."""
    bar: int = 3


def test_revision_id():
    """Check if there is a `revision_id` slipping into the schema."""
    schema = Foo.schema()
    assert "revision_id" not in schema["properties"]

    # check that the document has not been initialized, 
    # as otherwise the `revision_id` is normally gone from the schema.
    with pytest.raises(CollecctionWasNotInitialized):
        Foo.get_settings()
```

roman-right:
Hi! Sorry for the delay again. Thank you! I'll pick this up by this weekend

roman-right:
Ah, I see now. It covers the case when the model was not initialized. I tried to reproduce this with inited models and could not

roman-right:
Thank you for the PR! Merged.
It will be published in the next releases (beta or normal) today-tomorrow

ivan-gj:
Thank you for merging it! 
Happy to contribute to the project.
Looking forward to the upcoming release.

PR #486, opened 2023-02-08
Dnt create state on init for docs with custom id types

roman-right:


Luc1412:
Tested it, works fine. 👍 Thanks for fixing it

PR #502, opened 2023-03-14
Prevent the models returned from find_all to be modified in the middle of modifying

harris:
- do not execute the migration while collecting the input and output documents. 
- async gather so each batch could potentially run in parallel and not block each other. 
- verified it properly fixed the issues of any collections with size larger than 10,000 records.

### DIFF
--- a/beanie/migrations/controllers/iterative.py
+++ b/beanie/migrations/controllers/iterative.py
@@ -4,6 +4,7 @@ from typing import Type, Optional, Union, List
 from beanie.migrations.utils import update_dict
 from beanie.migrations.controllers.base import BaseMigrationController
 from beanie.odm.documents import Document
+import asyncio
 
 
 class DummyOutput:
@@ -90,6 +91,7 @@ def iterative_migration(
 
         async def run(self, session):
             output_documents = []
+            all_migration_ops = []
             async for input_document in self.input_document_model.find_all():
                 output = DummyOutput()
                 function_kwargs = {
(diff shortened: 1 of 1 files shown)


beanie/migrations/controllers/iterative.py
```diff
@@ -102,19 +104,22 @@ async def run(self, session):
...
-                    output_dict
-                )
+                    output_dict)
                 output_documents.append(output_document)
 
                 if len(output_documents) == self.batch_size:
-                    await self.output_document_model.replace_many(
-                        documents=output_documents
+                    all_migration_ops.append(
+                        self.output_document_model.replace_many(
```
jspiro, line 112: Suggest we wrap this promise with a print statement for each batch.

beanie/migrations/controllers/iterative.py
```diff
@@ -102,19 +104,22 @@ async def run(self, session):
...
+                            documents=output_documents
+                        )
                     )
                     output_documents = []
 
             if output_documents:
-                await self.output_document_model.replace_many(
-                    documents=output_documents
+                all_migration_ops.append(
+                    self.output_document_model.replace_many(
```
jspiro, line 120: Same.

beanie/migrations/controllers/iterative.py
```diff
@@ -102,19 +104,22 @@ async def run(self, session):
...
+                        documents=output_documents)
                 )
+            await asyncio.gather(*all_migration_ops)
```
jspiro, line 123: Or we can do it here and catch/rethrow errors. E.g.,
```
for op in asyncio.as_completed(all_migration_ops):
    try:
        print('Progress...')
        results = await op
    except Exception as e:
        print('Error:', e)
        throw e
```

jspiro:
I'd rename this to `prevent migrations from never completing`

harris:
@roman-right do you think you could take a look at this?  We actually had a data loss due to this bug in migration.

roman-right:
Merged. Thank you, guys!
It will be published in the next beta or normal release today or tomorrow (it depends on how many PRs will be included)

PR #517, opened 2023-03-31
kwargs arguments for ElemMatch operator

roman-right:


roman-right:
Hey @mikeckennedy,

Could you please play with this? It supports the next syntax currently: https://github.com/roman-right/beanie/pull/517/files#diff-ec52cedf58caab62054d0252af5d7960015cd4300d546d0125d51f6c67924c28R20-R28

I see, that it doesn't support deeply nested objects, but it supports the first layer at least.
Making embedded documents with attributes patched on the class level without additional initialization got more tricky than I expected. I still work on this direction too, but hacking it will probably take more time.

PR #546, opened 2023-04-22
Fix Table of Contents Not Showing All Classes

kschniedergers:
Noticed some of the table of contents on pages like https://beanie-odm.dev/api-documentation/operators/find were broken and it made it pretty difficult to look through.

TBH i think pydoc-markdown probably has a bug, as it feels like a workaround to do what I did, but putting anything to level 1 seems to make it not be seen for the table of contents.

Also my autolinter got rid of some whitespace


roman-right:
Hi! Thank you for the PR. I'll check it and merge tomorrow

PR #552, opened 2023-05-03
Feat: Add box operator

akriese:
Hi, 
I am currently using beanie for a project and was missing a builtin class for the `$box` operator.
So... here it is :)
Let me know, if this looks alright. 
If you dont like the `lower_left` and `upper_right` way of doing it, we could also do it simply with `coordinates` expecting a list with two tuples.

If you think, people might be interested in the other geospatial operators (center, etc.), I could also implement those. Just weren't part of my use case so far :)


roman-right:
Hi! This looks nice! Thank you :) I'll check everything and merge it tomorrow

PR #560, opened 2023-05-13
Reworked DB version handling, added verbose error for nested links incompatibility

Abashinos:
Added nested links incompatibility exception to query builder, along with a warning note in documentation ("Relations" tutorial section).

Added a fixture for easier database version parsing.

Moved DB version handling to a dedicated file, added compatibility utility methods in order to reduce hardcoded version matching.

Note: No unit tests have been added since the whole `tests/odm/test_relations.py` file is going to break if tested against MongoDB version that's older than 4.4. Since there is no test environment with an earlier MongoDB version, I don't find it necessary to add any workarounds there. A blanket pytest.skip might do the trick

### DIFF
--- a/.github/workflows/github-actions-tests.yml
+++ b/.github/workflows/github-actions-tests.yml
@@ -7,7 +7,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [ 3.7, 3.8, 3.9, 3.10.9, 3.11 ]
-        mongodb-version: [ 4.4, 5.0 ]
+        mongodb-version: [ 4.2, 4.4, 5.0 ]
         pydantic-version: [ 1.10.0 ]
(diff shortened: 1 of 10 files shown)


Abashinos:
Hold on, my tests seem redundant given that:
- House class already has nested links
- None of the test environments has MongoDB version older than 4.4

roman-right:
It looks like it doesn't pass all the tests.I'll pick this up after I finish with the Pydantic V2

Abashinos:
> It looks like it doesn't pass all the tests.I'll pick this up after I finish with the Pydantic V2

No worries, I would have already provided the fix if I didn't have my hands full as well...
We might come up with a patch for nested links in MongoDB 4.0/4.2 some time in the future, but that should likely go into a separate PR.

PR #569, opened 2023-05-23
fix(docs): documentation for ElemMatch

akriese:
The version with just matching on an array of values isn't possible. elemMatch needs an object (dict) as correctly specified in the type annotation.
I guess, the old version of the docstring was copy-pasted from the Any operator above :)


roman-right:
Good catch! Thank you

Issue #594, opened 2023-06-19
[BUG] generated openapi schema does not work with openapi-generator


alex-linx:
**Describe the bug**
I am building a FAST API app using beanie models.
I want to generate a typescript client for the api, using openapi-generator.
The FAST API app works ok on its own, however when I try to generate a client it says that there is a problem with the openapi spec.
It seems that the issue is with the examples field on the PydanticObjectId model

**To Reproduce**
```python
import json

from beanie import Document
from fastapi import FastAPI


class Item(Document):
    name: str


app = FastAPI()


@app.get("/", response_model=Item)
def get_item() -> Item:
    return Item(name="Foo")


if __name__ == "__main__":
    with open("openapi.json", "w") as f:
        json.dump(app.openapi(), f)
```

Run in the terminal
```
openapi-generator generate -i openapi.json -g typescript-axios
```

I am getting an error
```
Exception in thread "main" org.openapitools.codegen.SpecValidationException: There were issues with the specification. The option can be disabled via validateSpec (Maven/Gradle) or --skip-validate-spec (CLI).
 | Error count: 1, Warning count: 0
Errors: 
        -attribute components.schemas.Item.examples is unexpected

        at org.openapitools.codegen.config.CodegenConfigurator.toContext(CodegenConfigurator.java:620)
        at org.openapitools.codegen.config.CodegenConfigurator.toClientOptInput(CodegenConfigurator.java:647)
        at org.openapitools.codegen.cmd.Generate.execute(Generate.java:479)
        at org.openapitools.codegen.cmd.OpenApiGeneratorCommand.run(OpenApiGeneratorCommand.java:32)
        at org.openapitools.codegen.OpenAPIGenerator.main(OpenAPIGenerator.java:66)

```

**Additional context**
I am not sure whether the source of the issue is in the generator or in pydantic.
I looked at open api specification and it seems it does allow for an `example` field which is a single value example, or also `examples` - but from what I understand `examples` should be a dict, not a list.

**EDIT**
I have tried running the generated openapi.json through some online openapi validators and they also tell me that it is an invalid schema.
I think the following fix solves it:

```
class PydanticObjectId(ObjectId):
    """
    Object Id field. Compatible with Pydantic.
    """

    @classmethod
    def __modify_schema__(cls, field_schema):
        field_schema.update(
            type="string",
            example="5eb7cf5a86d9755df3a6c593",
        )

```

PR #595, opened 2023-06-19
Update fields.py - change examples to example

alex-linx:
Fix https://github.com/roman-right/beanie/issues/594


roman-right:
Hi! Thank you very much for the PR. It will be published by the end of this week

PR #620, opened 2023-07-18
Bump Pydantic | Beta 0

roman-right:


mkhorton:
Apologies if this is not the best place to ask, but is there a corresponding PR planned for `bunnet` for pydantic >= 2? Is it reasonable to expect `bunnet` to track `beanie` or is a lag or different level of maintenance expected? Thanks for the work on a great library.

roman-right:
Hi @mkhorton,

In 2-3 weeks after the release, I will publish the updated bunnet. At that time, I need to not only include the changes regarding Pydantic v2 but also integrate some additional features from beanie that are not yet present in bunnet.

mkhorton:
Thanks! Super helpful context

PR #628, opened 2023-07-22
DocumentSettings ignore extra fields Pydantic Validation

dynalz:
When setting the default pydantic Config to forbig extra fields:

```py
from pydantic import config, Extra
config.BaseConfig.extra = Extra.forbid
```

Beanie throws errors at document initialization due to the following: ValidationError: 4 validation errors for DocumentSettings __dict__
  extra fields not permitted (type=value_error.extra)
__doc__
  extra fields not permitted (type=value_error.extra)
__module__
  extra fields not permitted (type=value_error.extra)
__weakref__
  extra fields not permitted (type=value_error.extra)

The proposed commit fixes the issue. 
Not sure if ideal, or if there is any caveats with this solution

### DIFF
--- a/beanie/odm/settings/document.py
+++ b/beanie/odm/settings/document.py
@@ -1,6 +1,6 @@
 from typing import Optional, List
 
-from pydantic import Field
+from pydantic import Field, Extra
 
 from beanie.odm.fields import IndexModelField
 from beanie.odm.settings.base import ItemSettings
@@ -25,3 +25,4 @@ class DocumentSettings(ItemSettings):
 
     class Config:
         arbitrary_types_allowed = True
+        extra = Extra.ignore


roman-right:
Hi!

Sorry, I didn't get the use-case. Could you please explain it  bit?

dynalz:
So in our dev environment, we use Pydantic with more enforced rules and Dataclasses on production (for performance issues)

Thus in our dev environment we force all typing and all issues that mimic dataclasses, so that when it goes into production we have no issues. (This is for our service dataclasses, not related to beanie)

This is our use case:
```py
if SETTINGS.is_production() or TYPE_CHECKING:
    # production environment
    from dataclasses import dataclass
else:
    # dev environment
    # pydantic enforced configuration
    from pydantic.dataclasses import dataclass
    ...
    config.BaseConfig.extra = Extra.forbid
    config.BaseConfig.validate_all = True
    config.BaseConfig.validate_assignment = True
    config.BaseConfig.arbitrary_types_allowed = True
    ...
```

The `config.BaseConfig.extra = Extra.forbid` is so that it avoids anyone passing extra variables that are not meant for the dataclass as shown below:
```py
@dataclass
class Person:
    name: str

Person(**{"name": "John", "foo": "bear"}) # throws error if dataclasses module but not if pydantic module
```
this throws error as `@dataclass` comes from dataclasses module, but not if comes from pydantic.dataclasses module. Thus why the line `config.BaseConfig.extra = Extra.forbid` when using pydantic to mimic the dataclasses module behaviour.

However after we do the `config.BaseConfig.extra = Extra.forbid` (replacing the default `Extra.ignore` on pydantic) beanie throws at initializing `DocumentSettings` since it receives extra fields not defined on its model such as `__dict__`, `__doc__`, `__module__` and `__weakref__`

Effectively with this commit the behaviour should remain the same for beanie since the Pydantic Config `extra = Extra.ignore` is already the default and used on `DocumentSettings`


Let me know if this made sense

dynalz:
Let me know if some more clarification is needed

roman-right:
Hi. Sorry for the late reply. I had a lot of work with Pydantic V2 compatibility. If you still need this feature, could you please resolve the conflict and adapt this to work with Pydantic V2 (and still work with V1)?

Issue #653, opened 2023-08-10
merge_models does not merge list even if it does not include Link


roman-right:
### Discussed in https://github.com/roman-right/beanie/discussions/652

<div type='discussions-op-text'>

<sup>Originally posted by **tomohirohiratsuka** August 10, 2023</sup>
Hi, I'm wondering this is expected behavior or not.
When I update document with `update` method, it returns updated document only when the updated property is not list.
I found that inside update method it merging via `parsing.merge_models`.
`merge_models` looks like care about if the list includes Link or not but it does not merge even when no Links are found in the list.
Is this expected behavior?
I'd appreciate it, if anyone answer.
Thank you.

```python:documents.py
@wrap_with_actions(EventTypes.UPDATE)
    @save_state_after
    async def update(
        self,
        *args,
        ignore_revision: bool = False,
        session: Optional[ClientSession] = None,
        bulk_writer: Optional[BulkWriter] = None,
        skip_actions: Optional[List[Union[ActionDirections, str]]] = None,
        skip_sync: Optional[bool] = None,
        **pymongo_kwargs,
    ) -> DocType:
        """
        Partially update the document in the database

        :param args: *Union[dict, Mapping] - the modifications to apply.
        :param session: ClientSession - pymongo session.
        :param ignore_revision: bool - force update. Will update even if revision id is not the same, as stored
        :param bulk_writer: "BulkWriter" - Beanie bulk writer
        :param pymongo_kwargs: pymongo native parameters for update operation
        :return: None
        """

        arguments = list(args)

        if skip_sync is not None:
            raise DeprecationWarning(
                "skip_sync parameter is not supported. The document get synced always using atomic operation."
            )
        use_revision_id = self.get_settings().use_revision

        if self.id is not None:
            find_query: Dict[str, Any] = {"_id": self.id}
        else:
            find_query = {"_id": PydanticObjectId()}

        if use_revision_id and not ignore_revision:
            find_query["revision_id"] = self._previous_revision_id

        if use_revision_id:
            arguments.append(SetRevisionId(self.revision_id))
        try:
            result = await self.find_one(find_query).update(
                *arguments,
                session=session,
                response_type=UpdateResponse.NEW_DOCUMENT,
                bulk_writer=bulk_writer,
                **pymongo_kwargs,
            )
        except DuplicateKeyError:
            raise RevisionIdWasChanged
        if bulk_writer is None:
            if use_revision_id and not ignore_revision and result is None:
[...]

PR #655, opened 2023-08-11
fix: discussion#652

tomohirohiratsuka:
I've tried to fix #652 list merging behavior.
I'd appreciate if you check this fix.
Thank you.


tomohirohiratsuka:
This is solved on v1.22.1.
https://github.com/roman-right/beanie/issues/653

PR #657, opened 2023-08-14
Fix get_field_type & generalize extract_id_class

gsakkis:
- Return `outer_type_` instead of `type_` from `get_field_type` for pydantic v1
- Generalize `extract_id_class` to support arbitrary generic types

The combination of these two changes allow using a custom generic type for id.


roman-right:
Hi! Thank you for the PR. I think it works in 1.21. Could you please check?

gsakkis:
It doesn't, 1.21 was my starting point. 

MCVE:
```py
import beanie


class MyDocument(beanie.Document):
    id: int | None = None
    title: str


async def test_optional_id():
    client = AsyncIOMotorClient()
    await beanie.init_beanie(document_models=[MyDocument], database=client["test"])

    m = MyDocument(title="foo")
    assert m.id is None

    m.id = 42
    await m.insert()
    await client.drop_database("test")
```

Output:
```
E           ValueError: Unknown annotation: int | None
```

roman-right:
Thank you! Merged

PR #662, opened 2023-08-17
fix: AttributeError in odm.fields.Link.fetch_list()

pescn:
When all elements in links are documents, odm.fields.Link.fetch_list() will raise exceptions as: AttributeError: 'NoneType' object has no attribute 'find'. 
I fix it and add consistency check for document type and Link type.

### DIFF
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -292,9 +292,17 @@ class Link(Generic[T]):
                 else:
                     if document_class != link.document_class:
                         raise ValueError(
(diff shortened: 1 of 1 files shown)


roman-right:
Hi! Sorry, I didn't see this PR. I fixed this bug during my bug-fixing session. I did it a bit another way - I'm checking if the list with ids to fetch has any ids and only then start actual fetching. In this case it is impossible to face scenario, when document_class will stay none.

Thank you for the PR!

PR #665, opened 2023-08-20
Fix broken links

tsabsch:
Uses relative links instead of hardcoded URLs.


roman-right:
Thank you!

roman-right:
Thank you very much for your PR. Merged

PR #666, opened 2023-08-21
Fix: change the URL to discord in docs/development.md

jairoandrescastaneda:
The Discord URL in the link http://beanie-odm.dev/development/  points to a wrong URL, I fixed it by adding the correct URL in docs/development.md


roman-right:
Thank you for the fix!

PR #669, opened 2023-08-21
Fix | August 2023

roman-right:


beanie/odm/operators/find/logical.py
```diff
@@ -148,4 +148,19 @@ def __init__(self, expression: Mapping[str, Any]):
...
+            expression_key = list(self.expression.keys())[0]
+            if expression_key.startswith("$"):
+                raise AttributeError(
+                    "Not operator can not be used with operators"
+                )
+            value = self.expression[expression_key]
+            if isinstance(value, dict):
+                internal_key = list(value.keys())[0]
+                if internal_key.startswith("$"):
+                    return {expression_key: {"$not": value}}
```
gsakkis, line 161: @roman-right is the (omitted) `else` part here even valid, e.g. `{"$not": {"foo": 3}}`? All the examples I have seen with dict value are operator expressions, i.e. the condition here is always true. If that's the case this can be simplified. The added test doesn't cover this case so I can't tell for sure.

beanie/odm/queries/find.py
```diff
@@ -622,8 +622,13 @@ def build_aggregation_pipeline(self):
...
             self.document_model
         )
+        filter_query = self.get_filter_query()
+        if "$text" in filter_query:
+            text_query = filter_query["$text"]
+            aggregation_pipeline.insert(0, {"$match": {"$text": text_query}})
+            del filter_query["$text"]
 
-        aggregation_pipeline.append({"$match": self.get_filter_query()})
+        aggregation_pipeline.append({"$match": filter_query})
```
gsakkis, line 631: @roman-right shouldn't this logic be also applied to `AggregationQuery.get_aggregation_pipeline` ?

PR #673, opened 2023-08-23
Update precommit hooks & CI

gsakkis:
- Update precommit hooks
- Replace lint and mypy CI workflows with a precommit task in the test workflow
- Sort imports with `ruff`


roman-right:
Hi @gsakkis,
Thank you for the PR.
I think having a pre-commit check instead of separate MyPy and lint checks is good. However, it's better to have it in a separate GH action. It doesn't depend on Python or MongoDB versions, but it will run with each of them.

As often happens with formatter PRs, there are many conflicts. Could you please resolve them? Once you've done so, ping me on Discord before I merge anything else; this will help avoid creating more conflicts.

gsakkis:
Hi @roman-right, I updated the PR and moved the precommit to a separate job. Thanks!

roman-right:
Thank you very much for your PR. Merged

Issue #680, opened 2023-08-28
[BUG] Can't use a View if the user doesn't have full read privileges to all collections


marinashe:
**Describe the bug**
I'm trying to define a View using a user account that has limited read privileges. This causes the `list_collection_names` call to fail.

**To Reproduce**

1. Set up MongoDB with multiple collections and views.
2. Create a user account that only has read access to a specific view, which I'll refer to as "TestView," sourced from "Test" collection.

```python
class Test(Document):
    id: str

class TestView(View):
    id: str
    class Settings:
        name = 'TestView'
        source = 'Test'
        pipeline = [{"$match": {"id": "$id"}}]    

...
...
...

@app.on_event("startup")
async def startup_db_client():
    client = AsyncIOMotorClient(settings.DB_URL)
    await init_beanie(
        database=client[settings.DB_NAME],
        document_models=[TestView])
```
Run the app and it fails with the following traceback:

```
  File "../lib/python3.9/site-packages/starlette/routing.py", line 566, in __aenter__
    await self._router.startup()
  File "../lib/python3.9/site-packages/starlette/routing.py", line 654, in startup
    await handler()
  File "./backend/main.py", line 29, in startup_db_client
    await init_beanie(
  File "../lib/python3.9/site-packages/beanie/odm/utils/init.py", line 735, in init_beanie
    await Initializer(
  File "../lib/python3.9/site-packages/beanie/odm/utils/init.py", line 121, in __await__
    yield from self.init_class(model).__await__()
  File "../lib/python3.9/site-packages/beanie/odm/utils/init.py", line 705, in init_class
    await self.init_view(cls)
  File "../lib/python3.9/site-packages/beanie/odm/utils/init.py", line 645, in init_view
    collection_names = await self.database.list_collection_names()
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/concurrent/futures/thread.py", line 52, in run
    result = self.fn(*self.args, **self.kwargs)
  File "../lib/python3.9/site-packages/pymongo/database.py", line 1149, in list_collection_names
    return [result["name"] for result in self.list_collections(session=session, **kwargs)]
  File "../lib/python3.9/site-packages/pymongo/database.py", line 1101, in list_collections
    return self.__client._retryable_read(_cmd, read_pref, session)
  File "../lib/python3.9/site-packages/pymongo/_csot.py", line 108, in csot_wrapper
    return func(self, *args, **kwargs)
[...]

PR #681, opened 2023-08-28
fix | list_collection_names requires unnecessary privileges (#680)

marinashe:
This fixes the issue I encountered in #680.
The idea is to only list collections to which the user has access so that it will be possible to use a View with partially restricted users.


roman-right:
@marinashe Could you pls check GH checks?

roman-right:
Ah, no, don't worry. It looks like it's an old issue that the linter caught after the update. I'll merge this at the beginning of next week.

roman-right:
Merged. Thank you very much for your PR!

PR #694, opened 2023-09-07
feat(pydantic): add usage of pydantic.root_model if pydantic_v2

mathbeal:
code tested with pydantic 1.10 too.


roman-right:
Hi @mathbeal ,
Sorry, I didn't see your PR. It looks like this implemented in `1.22`. Could you please check?

PR #697, opened 2023-09-12
fix the document class name in comment

Pedram-Parsian:


roman-right:
Good catch! Thank you!

PR #699, opened 2023-09-12
Index support for custom pydantic types

aasay:
This allows custom pydantic types to used with indexes

### DIFF
--- a/beanie/__init__.py
+++ b/beanie/__init__.py
@@ -31,7 +31,7 @@ from beanie.odm.union_doc import UnionDoc
 from beanie.odm.utils.init import init_beanie
 from beanie.odm.views import View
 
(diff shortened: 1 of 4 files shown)


roman-right:
Hi @aasay,
You use walrus operator, which isn't supported by Python 3.7 (welcome to my reality :D). Can you please use less modern syntax? I plan to support 3.7 at least for this year.

aasay:
@roman-right,

Sorry about that.  I get very accustomed to using the newer versions of python and didn't even think about it.  It is updated now if you want to give it a try again.

roman-right:
Hi @aasay 
I'm sorry, it looks like I produced a conflict. Please let me know, if you can fix it. Alternatively I'll pick it up. Thank you for the PR!

PR #704, opened 2023-09-14
Added a feature that lets user create a Read Only Document

ahmarhafeez1:
I have added a feature that lets you create a document that is only able to read from DB and not modify it.

You can set read_only = True in Settings Object to make use of this feature by default it does not alter any other functionality of the ODM


requirements.txt
roman-right, line 1: I believe this one is not needed

beanie/odm/settings/base.py
```diff
@@ -20,7 +20,7 @@ class ItemSettings(BaseModel):
     union_doc: Optional[Type] = None
     union_doc_alias: Optional[str] = None
     class_id: str = "_class_id"
-
+    read_only: bool = False
```
roman-right, line 23: You have to choose or ItemSettings (it means it will be applied not only to Documents, but Views and etc) or DocumentSettings. Having this property in both is redundant

roman-right (changes requested):

roman-right:
Hi @ahmarhafeez1,

Thank you for the PR. 

I have a few comments. Please take a look there. 

Also, I'd like to discuss the feature in general. 

In this implementation, readonly documents can be changed with other methods, like class-level update:
```python
await ReadOnlySample.all().update(...)
```
This can be confusing, as users might think that all the readonly documents are protected. 

This can be avoided if you put a service field for such documents, like `_read_only=True`, in the database. Then, when you make updates or deletes, in the filter query part, you can add this condition to avoid such documents. And, probably, if a user needs to change this, you can add a special parameter to the changing operations. With this parameter set to true, it will change read-only docs.

What do you think?

roman-right:
> Yeah We can do that but I was thinking more along the lines of creating Read Only Documents for certain instances in your app where you need to read data if we added readonly property to DB it would permanently change those documents to readonly

I see. I still think it's better to make docs fully read-only, with the possibility of a force update using a special parameter. Based on my experience, users don't distinguish much between the logic of the instances and the logic of the classes. In this particular case, it can lead to serious mistakes.

PR #711, opened 2023-09-17
Preserve sort/skip/limit for aggregations

gsakkis:
Fix `FindMany.aggregate` to preseve any set sort/skip/limit stages.


roman-right:
Thank you for the PR! I'll check it very soon

roman-right:
Merged! It will be published by the end of this week together with others. Thank you!

PR #714, opened 2023-09-18
Fixed Link validation

iterlace:
In cases where you tried to save a `Document` with a field containing a `Link`/`DBRef` instead of the wrapped object itself, the `validate_on_save` hook would stop working, and return an error.

This was because the Link object was first serialized into a dict, and any further validation with `Link.validate` would always fail trying to deserialize a dict like `{"id": ..., "collection": ...}` as a real Document rather than a Link.

I've committed a simple solution, however, you can refine it to make it more elegant.

As my reserch has shown, there is no sense tweaking the way `Document.validate_self` validates the document. Instead, the easiest way would be to teach the `Link` to recognize its own schema.
This approach has the only obvious corner case: if the user creates a custom model with the same structure as Link, and then tries to wrap it with Link, he would get an unexpected behavior.


iterlace:
Ping?

roman-right:
Hi! Sorry for the delay. Thank you for the PR!
I'll check it by the end of this week

roman-right:
Merged! Thank you! It will be published by the end of this week with others.

PR #584, opened 2023-06-03
Refactor Encoder

gsakkis:
I took a stab at simplifying the `encoder` module a bit.

PR #730, opened 2023-10-04
fix: issue#728 pydantic_core._pydantic_core.Url object is not iterable

### LINKED ISSUE #728: [BUG] 'pydantic_core._pydantic_core.Url' object is not iterable
**Describe the bug**
Document.save()  throws error on pydnticV2.HttpUrl type fields

**To Reproduce**
```python
from beanie import Document, init_beanie
from pydantic import HttpUrl
import os

class Data(Document):
    link: HttpUrl

MONGO_URI = os.environ.get("MONGO_URI", "mongodb://localhost:27017/")


async def config_db():
    # Create Motor client
    client = AsyncIOMotorClient(MONGO_URI)

    # Init beanie with the Product document class
    await init_beanie(
        database=client["test"],
        document_models=[
            Data,
        ],
    )

await config_db()

data = Data(link="https://www.google.com")
await data.save()
```
error message:
```
File [venv/lib/python3.9/site-packages/beanie/odm/utils/encoder.py:244), in Encoder._encode(self, obj)
    243 try:
--> 244     data = dict(obj)
    245 except Exception as e:

TypeError: 'pydantic_core._pydantic_core.Url' object is not iterable

During handling of the above exception, another exception occurred:

TypeError                                 Traceback (most recent call last)
File [venv/lib/python3.9/site-packages/beanie/odm/utils/encoder.py:248](venv/lib/python3.9/site-packages/beanie/odm/utils/encoder.py:248), in Encoder._encode(self, obj)
    247 try:
--> 248     data = vars(obj)
    249 except Exception as e:

TypeError: vars() argument must have __dict__ attribute

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
     [24](line=23) await config_db()
     [26](line=25) data = Data(link="https://www.google.com/")
...
    250         errors.append(e)
--> 251         raise ValueError(errors)
    252 return self._encode(data)

ValueError: [TypeError("'pydantic_core._pydantic_core.Url' object is not iterable"), TypeError('vars() argument must have __dict__ attribute')]
```

**Expected behavior**
save runs without error out

**Additional context**

tomohirohiratsuka:
Hi,
To fix #728 which causing below errors, I've added `Url` type in encoder.
I'd appreciate if you check it.
Thank you.
```python
ValueError: [TypeError("'pydantic_core._pydantic_core.Url' object is not iterable"), TypeError('vars() argument must have __dict__ attribute')]
```


gsakkis:
@roman-right before fixing this please merge https://github.com/roman-right/beanie/pull/584. It's been open for 4 months and I already rebased and fixed the conflicts once, I probably won't bother again.

roman-right:
Hi @tomohirohiratsuka 
Sorry, I've produced a merge conflict here. Please let me know, if you can resolve it, otherwise I'll pick it up.

tomohirohiratsuka:
@roman-right 
Hi, I've resolved conflicts, please check it.
Thank you.

roman-right:
Hey @tomohirohiratsuka ,
It looks like a few checks failed. Could you please take a look?

beanie/odm/utils/encoder.py
```diff
@@ -1,155 +1,157 @@
...
+import ipaddress
+import operator
+import pathlib
+import re
+import uuid
+from typing import Any, Callable, Container, Iterable, Mapping, Optional, Tuple
+
+import bson
+import pydantic
+from pydantic_core import Url
```
roman-right, line 14: I think this one is not available for Pydantic V1. Can you please use a condition with `IS_PYDANTIC_V2`?

roman-right (changes requested):

tomohirohiratsuka:
@roman-right 
Hi, I've fixed the test case though it is a not perfectly clean since it has if condition inside the test.
Since pydantic v1, v2 url transform logic is changed like below.
I've confirmed beanie can handle both v1 same as before, v2 to fix bug.

Could you check changes?

```python
from pydantic import BaseModel, HttpUrl


class HttpUrlModel(BaseModel):
	url: HttpUrl

```

pydantic v1
```python
from main import HttpUrlModel


def test_debug():

	model = HttpUrlModel(url="https://www.google.com")
	assert model.url == "https://www.google.com" # True
	assert isinstance(model.url, str) # True
```

pydantic v2
```python
from pydantic_core import Url

from main import HttpUrlModel


def test_debug():

	model = HttpUrlModel(url="https://www.google.com")
	# assert model.url == "https://www.google.com" # This is not true in v2
	# assert isinstance(model.url, str) # This is not true in v2
	assert isinstance(model.url, Url) # This is true in v2, Url object is a new object in V2.

```

tests/odm/test_encoder.py
```diff
@@ -125,3 +128,30 @@ def test_keep_nulls_false():
     encoder = Encoder(keep_nulls=False, to_db=True)
     encoded_doc = encoder.encode(doc)
     assert encoded_doc == {"m": {"i": 10}}
+
+
+async def test_url():
+    # same logic from encoder.py IS_PYDANTIC_V2
+    is_pydantic_v2 = int(pydantic.VERSION.split(".")[0]) >= 2
```
roman-right, line 135: It is better to use IS_PYDANTIC_V2 variable here

tests/odm/test_encoder.py
```diff
@@ -125,3 +128,30 @@ def test_keep_nulls_false():
...
+
+        new_doc = await DocumentWithHttpUrlField.find_one(
+            DocumentWithHttpUrlField.id == doc.id
+        )
+
+        assert isinstance(new_doc.url_field, Url)
+        assert new_doc.url_field == doc.url_field
+    else:
+        doc = DocumentWithHttpUrlField(url_field="https://example.com")
+        assert isinstance(doc.url_field, str)
```
roman-right, line 149: Please check, if Pydantic V1 has a class for URL inherited from str. I'll check it too

roman-right:
@tomohirohiratsuka 
Having different tests for different Pydantic versions is ok. I left a few comments, please take a look

beanie/odm/utils/encoder.py
tomohirohiratsuka: Thanks, I'll change.

tests/odm/test_encoder.py
tomohirohiratsuka: Thanks, I'll use it.

tests/odm/test_encoder.py
tomohirohiratsuka: @roman-right 

Yes it has `AnyUrl` class.
Is it better to assert `AnyUrl` instead of `str` ?

```python
def test_debug():

	model = HttpUrlModel(url="https://www.google.com")
	assert model.url == "https://www.google.com"
	assert isinstance(model.url, AnyUrl) # True since HttpUrl(AnyHttpUrl(AnyUrl(str)))
```

roman-right: Yes, it is better, as it will display, that it was deserialized.
Hm. Will not `HttpUrl` work for both cases here then?

roman-right: Or maybe I'm missing the point of the test

tomohirohiratsuka: > Yes, it is better, as it will display, that it was deserialized.
> Hm. Will not HttpUrl work for both cases here then?

No it won't work since v1 use class inheritance but v2 uses subscripted generics for `HttpUrl`.

v2 networks.py
```python
AnyUrl = Url
"""A type that will accept any http or https URL."""
HttpUrl = Annotated[Url, UrlConstraints(max_length=2083, allowed_schemes=['http', 'https'])]
```
v2 _pydantic_core.py
```python
class Url(object):
```
v1 networks.py
```python
class AnyUrl(str):
```


But if we use `AnyUrl` as assertion it will pass both version without if condition though it hides the base class of instance.
```python
async def test_url():
    doc = DocumentWithHttpUrlField(url_field="https://example.com")
    assert isinstance(doc.url_field, AnyUrl)
    await doc.save()

    new_doc = await DocumentWithHttpUrlField.find_one(
        DocumentWithHttpUrlField.id == doc.id
    )

    assert isinstance(new_doc.url_field, AnyUrl)
    assert new_doc.url_field == doc.url_field
```

I think this depends on beanie's policy what we should assert exactly.

beanie/odm/utils/encoder.py
tomohirohiratsuka: To make it conditional, changing `Mapping `  to `MutableMapping` is acceptable ?

```python
DEFAULT_CUSTOM_ENCODERS: MutableMapping[type, SingleArgCallable] = {
    ipaddress.IPv4Address: str,
    ipaddress.IPv4Interface: str,
    ipaddress.IPv4Network: str,
    ipaddress.IPv6Address: str,
    ipaddress.IPv6Interface: str,
    ipaddress.IPv6Network: str,
    pathlib.PurePath: str,
    pydantic.SecretBytes: pydantic.SecretBytes.get_secret_value,
    pydantic.SecretStr: pydantic.SecretStr.get_secret_value,
    datetime.timedelta: operator.methodcaller("total_seconds"),
    enum.Enum: operator.attrgetter("value"),
    Link: operator.attrgetter("ref"),
    bytes: bson.Binary,
    decimal.Decimal: bson.Decimal128,
    uuid.UUID: bson.Binary.from_uuid,
    re.Pattern: bson.Regex.from_native,
}
if IS_PYDANTIC_V2:
    from pydantic_core import Url
    DEFAULT_CUSTOM_ENCODERS[Url] = str
```

tests/odm/test_encoder.py
roman-right: I think we have to be sure that it was encoded correctly. How it was decoded doesn't matter that much.
I'd check how the encoder encodes the document and would check if the output is correct. This way, we will be sure that it was stored in the db correctly. How it will be parsed is more Pydantic's responsibility. What do you think?

tomohirohiratsuka: Yes, I agree.
I've added a test for encoding.
Saving/Retrieving test case is remained to confirm it works in practice.

roman-right:
Merged! Thank you for the PR. It will be published tomorrow.

PR #741, opened 2023-10-14
Replace custom 'hidden=True' field attribute with builtin 'exclude=True'

gsakkis:
Replace the custom `hidden` Pydantic Field attribute with `exclude` and remove a bunch of code needed to support it.


roman-right:
Hi,
Thank you for the PR!
This is a breaking change. Please add a check and a warning log in case the `hidden` parameter was used.

gsakkis:
Restored backwards compatibility by transparently converting `hidden` field to `exclude` and logging a `DeprecationWarning`.

beanie/odm/documents.py
```diff
@@ -105,19 +103,23 @@
...
 
 
 def json_schema_extra(schema: Dict[str, Any], model: Type["Document"]) -> None:
-    props = {}
-    for k, v in schema.get("properties", {}).items():
-        if not v.get("hidden", False):
-            props[k] = v
-    schema["properties"] = props
+    # remove excluded fields from the json schema
+    properties = schema.get("properties")
```
roman-right, line 113: I don't remember how this call is implemented in Pydantic. Could you please check if the schema was not given as a mutable object here? If so, modifying it can lead to unexpected errors - it is better to clone it then.
I'll double-check this point from my side too.

roman-right (changes requested):

beanie/odm/documents.py
gsakkis: Not exactly sure what you mean, `schema` is a mutable object (dict) as shown in the annotation. The `json_schema_extra` callable returns None so it's only called for its side-effects, mutating schema. It would have no effect if the schema was cloned. 

The calling code in Pydantic v2 is here: https://github.com/pydantic/pydantic/blob/main/pydantic/json_schema.py#L1364-L1368

roman-right: Yep, you are right. This is expected behavior.

roman-right:
Merged. It will be published tomorrow

roman-right:
Thank you for the PR!

Issue #606, opened 2023-07-05
[BUG] `Text` query doesn't work with `fetch_links=True`


humbertoortuzar:
**Describe the bug**
Beanie @ 1.20.0
`fetch_links=True` works now with a `.count()` and `.aggregate(...)`, but...
When querying a document with links and a $text statement, the query fails because in the pipeline prefetched lookups are built before the `{$match: {$text: ... }}`. stage. Or at least when using `build_aggregation_pipeline`
```python
def build_aggregation_pipeline(self):
    aggregation_pipeline: List[Dict[str, Any]] = construct_lookup_queries(
        self.document_model
    )

    aggregation_pipeline.append({"$match": self.get_filter_query()})

    sort_pipeline = {"$sort": {i[0]: i[1] for i in self.sort_expressions}}
    if sort_pipeline["$sort"]:
        aggregation_pipeline.append(sort_pipeline)
    if self.skip_number != 0:
        aggregation_pipeline.append({"$skip": self.skip_number})
    if self.limit_number != 0:
        aggregation_pipeline.append({"$limit": self.limit_number})
    return aggregation_pipeline
```

So when you have a `.find(Text(search_query), fetch_links=True)` you get this error:
<img width="968" alt="image" src="https://github.com/roman-right/beanie/assets/53840891/eb1989ff-c512-4013-823a-83c151ed637e">

**To Reproduce**
```python
class ModelB(Document):
    some_field: str = Field(..., description="A field")
    other_field: Optional[str] = Field(None, description="Another field")

class ModelA(Document):
    name: Indexed(str, pymongo.TEXT) = Field(
        ..., description="Name of the ModelA instance."
    )

    model_b: Optional[Link[ModelB]] = Field(
        None, description="Reference to ModelA."
    )

    is_deleted: StrictBool = Field(False, description="Indicates if it's soft deleted")

query_result = await ModelA.find(Text("some text"), fetch_links=True).to_list()

```

**Expected behavior**
It should return a result instead of an error. To do this, `{$match: {$text: ...}}` queries should be included before the `construct_lookup_queries(self.document_model)`

**Additional context**
If anyone else is facing the same, what I'm doing right now is going inside the package to the function mentioned before and printing the stages needed to fetch your links and adding it to a `.aggregate()` query. If you also need to do a $match, you can add that stage to your pipeline.

In terms of the current beanie code, I was able to fix my case (multiple filters, fetch_links, and an aggregate pipeline) by doing this:
```python
[...]

PR #752, opened 2023-10-19
Fix aggregations with text queries

MrEarle:
This aims to fix the bug referred to in #606.

First time contributing here, so please let me know if I'm missing something or I'm not up to your standards 😁.

### DIFF
--- a/beanie/odm/queries/find.py
+++ b/beanie/odm/queries/find.py
@@ -45,7 +45,7 @@ from beanie.odm.queries.update import (
 )
 from beanie.odm.utils.dump import get_dict
 from beanie.odm.utils.encoder import Encoder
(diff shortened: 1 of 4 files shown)


roman-right:
Hi @MrEarle ,
Thank you for the PR! Could you please take a look at tests? It looks like python3.7 is not supported

MrEarle:
Oh, you're right! I'll fix it as soon as I have some time for it 👌🏼

MrEarle:
Fixed the tests. I was using `list` for typing instead of `List`.

roman-right:
Awesome! I'll play with it tomorrow and merge. Thank you!

roman-right (approved):

roman-right:
Good job! And I like how you avoided Python comparison operations, as they are overridden.

PR #754, opened 2023-10-19
Add support for Indexed custom pydantic fields

aasay:
This allows a custom pydantic class to be used with indexes


roman-right:
Thank you for the PR! Merged

PR #776, opened 2023-11-10
update model during save validation

roman-right:


hd10180:
Thanks for your quick response. i have try the fix. it works as expected。

PR #781, opened 2023-11-13
multiprocessing mode for init

roman-right:


adler-marmol:
@roman-right 
How does this mode work?
Do you have any documentation about it?
Can it help improve performance?

PR #786, opened 2023-11-17
Improve write performances

tdjaballah:
* Improve performance when a document has a lot of sub document


roman-right:
Hi @tdjaballah ,
Thank you for the PR! Merged. Will be published soon

PR #810, opened 2023-12-18
Update documents.py

CAPITAINMARVEL:
fix uniondoc DocType

### DIFF
--- a/beanie/odm/interfaces/find.py
+++ b/beanie/odm/interfaces/find.py
@@ -25,12 +25,10 @@ from beanie.odm.queries.find import FindMany, FindOne
 from beanie.odm.settings.base import ItemSettings
 
 if TYPE_CHECKING:
-    DocType = TypeVar("DocType", bound=Union["Document", "UnionDoc"])
-    from beanie.odm.documents import Document
-    from beanie.odm.views import View
+    # only used as type hint
+    DocType = TypeVar("DocType", bound=Union["Document", "UnionDoc"])  # type: ignore
 
 DocumentProjectionType = TypeVar("DocumentProjectionType", bound=BaseModel)
-FindType = TypeVar("FindType", bound=Union["Document", "View"])
 
 
 class FindInterface:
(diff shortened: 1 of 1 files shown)


beanie/odm/queries/find.py
```diff
@@ -807,6 +866,74 @@ def find_one(
...
+        Create Update with modifications query
+        and provide search criteria there
+
+        :param args: *Mapping[str,Any] - the modifications to apply.
+        :param session: Optional[ClientSession]
+        :param bulk_writer: Optional[BulkWriter]
+        :param response_type: UpdateResponse
+        :return: UpdateMany query
+        """
+        self.set_session(session)
```
roman-right, line 888: It looks unnecessary to keep the logic part inside the method marked with `@overload`. It will not be called at the runtime.

beanie/odm/queries/find.py
```diff
@@ -352,14 +352,14 @@ def sort(
...
-                str, Tuple[str, SortDirection], List[Tuple[str, SortDirection]]
+                Any, Tuple[Any, SortDirection], List[Tuple[Any, SortDirection]]
             ]
         ],
     ) -> "FindMany[FindQueryResultType]":
         """
         Add sort parameters
-        :param args: Union[str, Tuple[str, SortDirection],
-        List[Tuple[str, SortDirection]]] - A key or a tuple (key, direction)
+        :param args: Union[Any, Tuple[Any, SortDirection],
```
roman-right, line 361: I Any a good option? Probably it is possible to limit the number of possible types? Or there is something that blocks it?

roman-right (changes requested):
Thank you very much for the PR. Please take a look at the comments and at the pre-commit validation test.
Thank you!

roman-right:
Hey @CAPITAINMARVEL ,
Could you please take a look at the tests? 

Regarding the amount of commits - everything is ok. I use squash merging strategy - it will be 1 commit in the main branch in the end anyway :)

roman-right:
Hi @CAPITAINMARVEL ,
I'll fix the merge conflict and take a look at the error by the end of today

roman-right:
Hm, it looks like part of the problem has alternative solution there already. Could you please take a look? @CAPITAINMARVEL

PR #813, opened 2023-12-21
Add missing type signature to `Document` constructor

### LINKED ISSUE #812: [BUG] `Document` constructor fails type check for instantiation with Mypy in `strict` mode
**Describe the bug**

Beanie is missing complete type annotations on some objects, such as `Document`, which cause Mypy to fail when used.

**To Reproduce**

The following code:

```py
from beanie import Document

d = Document()
```

Throws a type check error when run with Mypy:

```
$ mypy main.py
main.py:3: error: Call to untyped function "Document" in typed context  [no-untyped-call]
Found 1 error in 1 file (checked 1 source file)
```

With the following Mypy configuration:

```toml
[tool.mypy]
strict = true

plugins = [
    "pydantic.mypy",
]
```

This config can be further simplified to

```toml
[tool.mypy]
disallow_untyped_calls = true

plugins = [
    "pydantic.mypy",
]
```

This seems to be because this line, the constructor is not fully typed, it is missing `-> None`:

https://github.com/roman-right/beanie/blob/5f03c6f6312fd0232505cc9bebf8ac1b22ca95ac/beanie/odm/documents.py#L189

**Expected behavior**

No Mypy type check errors when using Beanie, even with in `strict` mode.

**Additional context**

Reproduced on:

- Python 3.11.6
- Mypy 1.7.1
- Beanie 1.23.6

Somewhat related to

- #672
- #769

johnthagen:
Closes #812

### DIFF
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -186,7 +186,7 @@ class Document(
             self._previous_revision_id = self.revision_id
             self.revision_id = uuid4()
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
         super(Document, self).__init__(*args, **kwargs)
         self.get_motor_collection()
 


roman-right:
Thank you very much for the PR! Merged

PR #826, opened 2024-01-07
fix: count with text queries and links

MrEarle:
This PR aims to fix the `count` method when combining links and text queries.

Currently, the `count` method has the same issue as described in #606. This PR aims to implement the fix in #752 for the `count` method.


roman-right:
Hi @MrEarle ,
Thank you for the PR!

Issue #839, opened 2024-01-24
[BUG] update_many with aggregate set fields to a nested document not the exact expression value 


SolardiaX:
**Describe the bug**
When use find & update_many to replace array fields frist N elements with new elements, the command update the fields value to nest document not the expresssion calculated value

**To Reproduce**
```python
origin_children_ancestors = [ObjectId("65acdb2f025de11c4b6f4fa8"),ObjectId("65ad037edbe9cca114dcaf38")]
new_children_ancestors = [ObjectId("65acdb2f025de11c4b6f4fa8"),ObjectId("65ad037edbe9cca114dcaf38"), ObjectId(""65ae10c1652cea51d101c0c1"")]

await Categories.find({
    f"ancestors.{i}": val for i, val in enumerate(origin_children_ancestors)
}).update_many({
    "$set": {
        "ancestors": {
            "$concatArrays": [
                new_children_ancestors,
                {"$slice": ["$ancestors", len(origin_children_ancestors), {"$size": "$ancestors"}]}
            ]
        }
    }
})
```

**Expected behavior**
"ancestors" field should be updated starts with "new_children_ancestors" array.

**Additional context**
"ancestors" is wrong updated to a dict {
  "$concatArrays": [
      new_children_ancestors,
      {"$slice": ["$ancestors", len(origin_children_ancestors), {"$size": "$ancestors"}]}
]}

SolardiaX:
I think we should change the *args: Mapping[str, Any] signture of the update_many method to *args: Mapping[str, Any] | Sequence[Mapping[str, Any]] and used as:
```python

await Categories.find({
    f"ancestors.{i}": val for i, val in enumerate(origin_children_ancestors)
}).update_many([{
    "$set": {
        "ancestors": {
            "$concatArrays": [
                new_children_ancestors,
                {"$slice": ["$ancestors", len(origin_children_ancestors), {"$size": "$ancestors"}]}
            ]
        }
    }
}])
```

Currently, this script can be executed as expect but with issue hints in IDE.

roman-right:
Thank you for the catch. I'll pick this up on the next bug fixing session

PR #840, opened 2024-01-24
support pipeline in update

SolardiaX:
see the issue [#839](https://github.com/roman-right/beanie/issues/839)


roman-right:
Hi @SolardiaX ,
It looks like some of the python version don't support features you used. Could you please check?

PR #861, opened 2024-02-15
handle TypeError in validator of PydanticObjectId

grthr:
the initializer of bson.ObjectId might raise `TypeError(f"id must be an instance of (bytes, str, ObjectId), not {type(oid)}")`.

Pydantic validators should raise `ValueError`s or `AssertionError`s instead.


roman-right:
Thank you for the PR! It will be published soon

PR #866, opened 2024-02-18
Add `with_children: bool` to `delete_all()`, to allow calling delete_all on an `is_root=True` base class, to delete any subclass rows as well.

TheBloke:
A simple change: adds `with_children: bool` to `delete_all()`, which gets passed through to `find_all()`.

When there is an inheritance tree, this allows deleting all documents in a collection using the base class in one simple operation.

### DIFF
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -906,6 +906,7 @@ class Document(
         cls,
         session: Optional[ClientSession] = None,
         bulk_writer: Optional[BulkWriter] = None,
+        with_children: bool = False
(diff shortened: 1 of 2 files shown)


roman-right:
Hi! Thank you for the pr. Please check the tests

TheBloke:
Sorry, fixed

roman-right:
@TheBloke thank you for the PR. It looks good. Could you please add a test case for this?

roman-right:
Thank you @CAPITAINMARVEL ,
I'll add your tests to the branch this week.

PR #871, opened 2024-02-21
Python Exception Chaining

roshanr10:
Currently a few exceptions are lost due to the way Beanie handles the try/except & raise approach – using exception chaining / re-raising the same exception ensures the appropriate values can be used for debugging downstream.

Especially with `DuplicateKeyError`, it's incredibly unclear to the user when they get a `RevisionIdWasChanged` instead with no way to trace it back to a `DuplicateKeyError`.

### DIFF
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -513,11 +513,11 @@ class Document(
                 session=session,
                 bulk_writer=bulk_writer,
             )
-        except DocumentNotFound:
+        except DocumentNotFound as e:
             if use_revision_id and not ignore_revision:
-                raise RevisionIdWasChanged
+                raise RevisionIdWasChanged from e
             else:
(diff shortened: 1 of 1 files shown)


roman-right:
Hi @roshanr10 ,
Thank you very much for the PR! Could you please add a test for this case?

roshanr10:
Hi Roman, thanks for the update, following up on this – can you please provide more insight into why DuplicateKeyError is being caught? As far as I've experienced, there is no unique ID index on the revision id, so it would never be caught in normal usage.

this is what I've done to get information from the underlying DuplicateKeyError but not sure why it's intercepted
```python
except beanie.exceptions.RevisionIdWasChanged as e:
    if e.__context__.details["keyPattern"]["<YOUR_KEY_PATTERN"] == 1:
        # HANDLE
```

Issue #884, opened 2024-02-25
[BUG] Argument of type `Callable` to `json_schema_extra` raises `AttributeError`


valentinoli:
**Describe the bug**
When I pass a `Callable` to the `json_schema_extra` argument of `pydantic.Field()`, I get the following error:

```
  File "/home/xxxx/project/.venv/lib/python3.11/site-packages/beanie/odm/utils/init.py", line 754, in init_beanie
    await Initializer(
  File "/home/xxxx/project/.venv/lib/python3.11/site-packages/beanie/odm/utils/init.py", line 129, in __await__
    yield from self.init_class(model).__await__()
  File "/home/xxxx/project/.venv/lib/python3.11/site-packages/beanie/odm/utils/init.py", line 717, in init_class
    await self.init_document(cls)
  File "/home/xxxx/project/.venv/lib/python3.11/site-packages/beanie/odm/utils/init.py", line 573, in init_document
    self.init_document_fields(cls)
  File "/home/xxxx/project/.venv/lib/python3.11/site-packages/beanie/odm/utils/init.py", line 385, in init_document_fields
    cls.check_hidden_fields()
  File "/home/xxxx/project/.venv/lib/python3.11/site-packages/beanie/odm/documents.py", line 1055, in check_hidden_fields
    hidden_fields = [
                    ^
  File "/home/xxxx/project/.venv/lib/python3.11/site-packages/beanie/odm/documents.py", line 1058, in <listcomp>
    if get_extra_field_info(field, "hidden") is True
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xxxx/project/.venv/lib/python3.11/site-packages/beanie/odm/utils/pydantic.py", line 45, in get_extra_field_info
    return field.json_schema_extra.get(parameter)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'function' object has no attribute 'get'
```

See [Pydantic Documentation](https://docs.pydantic.dev/latest/concepts/json_schema/#using-json_schema_extra-with-a-callable)

**To Reproduce**
```python
from pydantic import Field
from beanie import Document

def pop_default(s):
    s.pop('default')

class Model(Document):
    a: int = Field(default=1, json_schema_extra=pop_default)
```

**Expected behavior**
This should work and be compatible with Pydantic.

roman-right:
Hi! Thank you for the catch.

PR #896, opened 2024-03-12
fix: allow json_schema_extra as Callable (#884)

valentinoli:
Simply check if `field.json_schema_extra` is a `dict`. That way `None` is returned when it is a `Callable`. It seems too complicated to support `Callable`, it would require calling the function with the schema as an argument. I would assume that function call is only done internally in Pydantic so there is not a clear way to do this and retrieve the modified schema after calling the function. So I propose not to support that for now. This means for instance that adding `original_field` to the JSON schema of a `beanie.BackLink` field in a callable would not be supported...

See #884

### DIFF
--- a/beanie/odm/utils/pydantic.py
+++ b/beanie/odm/utils/pydantic.py
@@ -41,7 +41,7 @@ def parse_model(model_type: Type[BaseModel], data: Any):
 
 def get_extra_field_info(field, parameter: str):
     if IS_PYDANTIC_V2:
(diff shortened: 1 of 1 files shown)


roman-right:
Hi @valentinoli ,
Looks good. Thank you. I plan to merge the PR at this weekend

valentinoli:
Great, thank you :)

roman-right:
Merged. It will be published today

Issue #676, opened 2023-08-25
Beanie 1.21 still triggers many deprecation warnings with Pydantic v2


WarpedPixel:
**Describe the bug**
There are many Pydantic v2 deprecation warnings triggered by beanie code.

**To Reproduce**
run python code using beanie 1.21 with `-W error` to stop at the first warning.

Example:
```
../../../Library/Caches/pypoetry/virtualenvs/companion-mgXvcyvA-py3.11/lib/python3.11/site-packages/pydantic/_internal/_config.py:219: in prepare_config
    warnings.warn(DEPRECATION_MESSAGE, DeprecationWarning)
E   pydantic.warnings.PydanticDeprecatedSince20: Support for class-based `config` is deprecated, use ConfigDict instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.3/migration/
```

Triggered by this code in `beanie/odm/bulk.py`:

```python
class Operation(BaseModel):
    operation: Union[
        Type[InsertOne],
        Type[DeleteOne],
        Type[DeleteMany],
        Type[ReplaceOne],
        Type[UpdateOne],
        Type[UpdateMany],
    ]
    first_query: Mapping[str, Any]
    second_query: Optional[Dict[str, Any]] = None
    pymongo_kwargs: Dict[str, Any] = Field(default_factory=dict)
    object_class: Type

    class Config:
        arbitrary_types_allowed = True
```

**Expected behavior**
Clean run with no deprecations.

**Additional context**
I see primarily these 4 types of deprecation warnings:

```
  /Users/bcs/Library/Caches/pypoetry/virtualenvs/companion-mgXvcyvA-py3.11/lib/python3.11/site-packages/beanie/odm/utils/init.py:386: PydanticDeprecatedSince20: The `__fields__` attribute is deprecated, use `model_fields` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.3/migration/
[...]

PR #897, opened 2024-03-15
Update deprecated call of general_plain_validator_function (#676)

dslemusp:
https://docs.pydantic.dev/latest/api/pydantic_core_schema/#pydantic_core.core_schema.with_info_plain_validator_function

### DIFF
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -147,7 +147,7 @@ class PydanticObjectId(ObjectId):
             cls, source_type: Any, handler: GetCoreSchemaHandler
         ) -> CoreSchema:  # type: ignore
             return core_schema.json_or_python_schema(
-                python_schema=core_schema.general_plain_validator_function(
+                python_schema=core_schema.with_info_plain_validator_function(
                     cls.validate
                 ),
                 json_schema=str_schema(),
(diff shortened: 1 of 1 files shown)


hasB4K:
I would love this to merged! Most of my pytests warning on a project of mine are actually caused by this.

@roman-right, FYI, the deprecation was done by this PR: https://github.com/pydantic/pydantic-core/pull/980
And I think this PR can be merged like this since others projects just did exactly what @dslemusp is doing in this PR (a simple function replacement, see: https://github.com/vitalik/django-ninja/pull/922)

roman-right:
Hi @hasB4K ,

Thank you for following up. I'll pick it up at the beginning of the next week.

beanie/odm/fields.py
```diff
@@ -414,7 +414,7 @@ def __get_pydantic_core_schema__(
                     lambda instance: cls.serialize(instance)  # type: ignore
                 ),
             )
-            return core_schema.general_plain_validator_function(
+            return core_schema.with_info_plain_validator_function(
                 cls.build_validation(handler, source_type)
             )
```
hasB4K, line 419: I was reading the original code, and my IDE seems to complain that this code is unreachable.

dslemusp: Yeah that was also my case but as my PR is only related to deprecation I only changed the deprecated calls

roman-right:
It looks good. I plan to have it merged at this weekend

rmariano (approved):

roman-right:
Merged. It will be published today

Issue #749, opened 2023-10-16
[BUG] Optional[BackLinks] return beanie.odm.fields.BackLink object


Nicialy:
**Describe the bug**
BackLink with None Create return dont None

**To Reproduce**
```python

class Sprint(Document, SprintCreation):
    id: UUID = Field(default_factory=uuid4)
    issues: List[Link["Issue"]] = Field(default_factory=list)

class Issue(Document, IssueBase):
    id: UUID = Field(default_factory=uuid4)
    creation_date: datetime = Field(default_factory=datetime.now)
    sprint: Optional[BackLink["Sprint"]] = Field(default=None,original_field="issues", exclude=True)

issue = Issue(**issue_creation.model_dump(),sprint=None)
issue = await Issue.find_one(Issue.id == issue_id, fetch_links=True)
if issue.sprint == None or issue_creation.sprint_id != issue.sprint.id:
            issue.sprint.issues.remove(issue.id)

```

**Expected behavior**
 I want issue.sprint == None true

![image](https://github.com/roman-right/beanie/assets/43730314/885ef021-9a1d-4fe9-bb0e-93bee9917571)
![image](https://github.com/roman-right/beanie/assets/43730314/08511ea9-be84-4712-88ba-4d52709eac1e)

roman-right:
Hi! Could you please provide on-module example of this behavior which I can run locally to check, what is happening?

Nasochec:
Example:
[main.py.txt](https://github.com/roman-right/beanie/files/13266539/main.py.txt)

Nicialy:
```python
from motor.motor_asyncio import AsyncIOMotorClient

from typing import Optional
from pydantic import Field
import asyncio
from beanie import init_beanie,Document,BackLink,Link
from beanie.odm.fields import PydanticObjectId

class Country(Document):
    id: PydanticObjectId = Field(default_factory=PydanticObjectId)
    cities: list[Link["City"]] = Field(default_factory=list)

class City(Document):
    id: PydanticObjectId = Field(default_factory=PydanticObjectId)
    country: Optional[BackLink["Country"]] = Field(default=None, original_field="cities", exclude=True)

__beanie_models__=[Country,City]

async def fillSomeData(): 
    session:AsyncIOMotorClient  = AsyncIOMotorClient("mongodb://localhost:27017")
    await init_beanie(database=session.example, document_models=__beanie_models__)
  
    cityOk = City()
    await cityOk.create()
[...]

PR #904, opened 2024-03-27
Update Relations docs to indicate that backlinks are virtual.

JBorrow:
The BackLink docs did not include information about them being virtual, causing confusion to me. I added this documentation following the comment thread here: https://github.com/roman-right/beanie/issues/749.


roman-right:
Thank you for the update, @JBorrow

roman-right:
Merged

PR #907, opened 2024-03-28
improving and fix type hint

CAPITAINMARVEL:
to be continued ...

PR #909, opened 2024-04-03
Fixing type hint and correct return for asynchronous update like (Document.find().inc()

CAPITAINMARVEL:


tristandeborde:
@bedlamzd I'm pretty sure this is a duplicate of #907 for which you already called out the author (https://github.com/roman-right/beanie/pull/907#issuecomment-2033351625)
@CAPITAINMARVEL , this problem is already going to be fixed with https://github.com/roman-right/beanie/pull/886

roman-right:
Hi @CAPITAINMARVEL , could you please check the red tests and there are a few merge conflicts now.
Thank you very much for the PR!

PR #912, opened 2024-04-07
added time encoder

07pepa:
I have added time encoded however i suspect there may be issue in pydantic ( speed date) itself see https://github.com/pydantic/speedate/issues/62

### DIFF
--- a/beanie/odm/utils/encoder.py
+++ b/beanie/odm/utils/encoder.py
@@ -26,6 +26,65 @@ import beanie
 from beanie.odm.fields import Link, LinkTypes
 from beanie.odm.utils.pydantic import IS_PYDANTIC_V2, get_model_fields
 
+UNIX_TIME_START = datetime.datetime(1970, 1, 1, 0, 0, 0, 0)
+
+
+def _compatct_time_encoder(time: datetime.time) -> str:
+    # re-implementation of speed date in python
+    # https://github.com/pydantic/speedate/blob/aad1d9117a05618ae883e20ae179c582eb998eeb /src/time.rs#L41 and
+    # pydantic https://github.com/pydantic/pydantic-core/blob/f636403c2e8169bcbb94aa71a0727d76076f7e6e/src/input
+    # /datetime.rs#L177
+    if time.microsecond != 0:
+        time_string = "{:02d}:{:02d}:{:02d}.{:06d}".format(
+            time.hour, time.minute, time.second, time.microsecond
+        )
+    elif time.second != 0:
+        time_string = "{:02d}:{:02d}:{:02d}".format(
+            time.hour, time.minute, time.second
+        )
+    else:
+        time_string = "{:02d}:{:02d}".format(time.hour, time.minute)
+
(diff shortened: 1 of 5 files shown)


roman-right:
Hi @07pepa ,
THank you for the PR. It looks interesting. I'll check it out this weekend

beanie/odm/utils/encoder.py
```diff
@@ -26,6 +26,58 @@
...
+            time.hour, time.minute, time.second
+        )
+    else:
+        time_string = "{:02d}:{:02d}".format(time.hour, time.minute)
+
+    tz_offset = None
+    if time.tzinfo is not None:  # has timezone
+        # check if the offset is fixed no matter the datetime
+        offset_delta = time.tzinfo.utcoffset(None)
+        if offset_delta is None:  # is ambiguous resolve base on current time
```
pepe-rtmlab, line 50: i am not shure if this is right... 
this may l change serialization based on date... 

This is caused when there is two variance of timezone  (two timedeltas) like Europe/Prague having summer and winter variant(DST) and its not conclusive what is right

alternatives:

1. throw error that timezone is inconclusive 
2. serialize to particular date (like 1.1.1970) but this ventures to defining a standard

tests/odm/test_encoder.py
```diff
@@ -169,3 +181,87 @@ async def test_dict_with_complex_key():
...
+        time(12, 4, 5, 123456, tzinfo=ZoneInfo("Europe/Prague")),
+        time(12, 4, 5, 123456, tzinfo=ZoneInfo("Europe/Prague"), fold=1),
+    ],
+)
+async def test_encode_time_with_tz(test_time: time):
+    await inner_test_time(test_time)
+
+
+if has_zone_info:
+    tz = list(available_timezones())
```
pepe-rtmlab, line 253: maybe add cross test between pytz and zone info?

roman-right:
Hi @07pepa ,
Could you please check the red tests?

07pepa:
now is green https://github.com/07pepa/beanie/actions/runs/8930039487/job/24529213097

07pepa:
lets wait how pydantic will resolve this  here https://github.com/pydantic/pydantic/issues/9571

roman-right:
> lets wait how pydantic will resolve this here
Yes, let's wait. THank you!

Riverfount (approved):
I read the code and I think everything is fine. My question is why GitHub Actions doesn't run the jobs. But, if the tests are passed, I approve this PR.

07pepa:
@Riverfount  there is preaty nasty bug in  https://github.com/pydantic/pydantic/issues/9571 we should wait how they resolve it IMHO it would be bad if implementation would differ from theirs  and it currently does and bypasses pydantic

07pepa:
@Riverfount can you remove you approve ? i know this is not usual request... but... in this case is better to wait

Riverfount (commented):
Waiting pydantic bug fix.

PR #929, opened 2024-05-06
Type hint overall improving

CAPITAINMARVEL:
Fix correct return type for query update operation, fix type hint for sort, correct return type for document update and removed awaitable making it not able to be used in asyncio task

### DIFF
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -694,7 +694,7 @@ class Document(
         skip_actions: Optional[List[Union[ActionDirections, str]]] = None,
         skip_sync: Optional[bool] = None,
         **pymongo_kwargs: Any,
-    ) -> DocType:
+    ):
         """
         Partially update the document in the database
 
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -5,7 +5,6 @@ from enum import Enum
 from typing import (
     TYPE_CHECKING,
     Any,
-    Awaitable,
     Callable,
     ClassVar,
     Coroutine,
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -761,7 +761,7 @@ class Document(
         )
 
     def set(
-        self: DocType,
+        self,
         expression: Dict[Union[ExpressionField, str], Any],
         session: Optional[AsyncIOMotorClientSession] = None,
         bulk_writer: Optional[BulkWriter] = None,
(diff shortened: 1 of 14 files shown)


beanie/odm/documents.py
```diff
@@ -689,17 +688,17 @@ async def update(
...
         ignore_revision: bool = False,
-        session: Optional[ClientSession] = None,
+        session: Optional[AsyncIOMotorClientSession] = None,
         bulk_writer: Optional[BulkWriter] = None,
         skip_actions: Optional[List[Union[ActionDirections, str]]] = None,
         skip_sync: Optional[bool] = None,
-        **pymongo_kwargs,
-    ) -> DocType:
+        **pymongo_kwargs: Any,
+    ):
```
alm0ra, line 696: Why you remove return type? 
If you using mypy as a type checker you will facing Error

alm0ra: If it doesn't return any thing 
You can using '-> None'

beanie/odm/documents.py
```diff
@@ -1247,19 +1246,19 @@ async def delete(
     @classmethod
     def find_many_in_all(  # type: ignore
         cls: Type[FindType],
-        *args: Union[Mapping[str, Any], bool],
+        *args: Union[Mapping[Any, Any], bool],
```
alm0ra, line 1249: Why you all *args to Any?
Why it was mpping[str,any]

alm0ra:
Hi @CAPITAINMARVEL ,
How did you test this types?
Did you use type checker?

adeelsohailahmed:
@CAPITAINMARVEL, please resolve the conflicts.

beanie/odm/interfaces/find.py
```diff
@@ -39,7 +44,7 @@ class FindInterface:
     _find_one_query_class: ClassVar[Type] = FindOne
     _find_many_query_class: ClassVar[Type] = FindMany
 
-    _inheritance_inited: bool = False
+    _inheritance_inited: bool
```
adeelsohailahmed, line 47: Why was this default to `False` removed?

adeelsohailahmed (changes requested):
I think it'd be easier for both you, and the PR reviewers, if you were to break your big PR into smaller PRs, with each PR dealing with a specific module.

Could you please break this PR into multiple, smaller PRs?

PR #932, opened 2024-05-10
Fix example in multi-model.md

gianpaj:
```python
docs = await Two.all().to_list()
```

returns documents of class `Two`


roman-right:
Thank you for the PR!

PR #935, opened 2024-05-21
Fix: Set default value in FindInterface._inheritance_inited to avoid …

Robert-Nogueira:
…AttributeError in _add_class_id_filter method

Fixed error raised when in "FindInterface._add_class_id_filter" method (line 458) when a attempt to acess the value from "FindInterface._inheritance_inited"


roman-right (approved):
Looks good. Thank you

PR #938, opened 2024-05-23
[Feature] Adding documents with session

tistaharahap:
Instead of passing the `session` objects in document read/write methods, use a `ContextVar` to hold a session for documents. This makes it a lot simpler if this is built in with the `Document` object. This PR tries to do that.

Also added a page in the docs site to describe how to use it.


roman-right:
Hi @tistaharahap , please take a look at the red tests.
And thank you for the PR!

PR #942, opened 2024-06-06
feat(skip_index): possibility added to skip index actions

jorma16:
Just added the possibility to avoid index operations in init, like someone ask here: https://github.com/BeanieODM/beanie/discussions/924

And other people workaround here: https://github.com/BeanieODM/beanie/discussions/608

Please, take care of this, I really need this and I think that is very interesting for read only applications.


roman-right:
Hi @jorma16 ,
Thank you very much for the PR. THis feature is so reasonable. Could you please add a test for it? It could check if indexes were not created in the db.

jorma16:
Hi! I couldn't check this PR until today, sorry for the delay.

I have added the tests that you requested, thanks in advance!

adeelsohailahmed (approved):
Thank you for your contribution!

staticxterm (approved):

Riverfount (approved):
It's ok for me.

PR #953, opened 2024-06-20
Add UUID4 support for InspectionError

### LINKED ISSUE #838: [BUG] UUIDs are not supported as ID types in Document.inspect_collection
**Describe the bug**
MongoDB natively supports UUIDs as object IDs (see https://stackoverflow.com/questions/28895067/using-uuids-instead-of-objectids-in-mongodb). However when using Document.inspect_collection with ids that are UUIDs, it gives an error:

```
 TypeError: id must be an instance of (bytes, str, ObjectId), not <class 'uuid.UUID'>
```

**To Reproduce**
```python
from beanie import Document
from pydantic import UUID4

class MyCol(Document):
  _id: UUID4
```

nikstuckenbrock:
Fixes #838.

### DIFF
--- a/beanie/odm/models.py
+++ b/beanie/odm/models.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Union
 
 from pydantic import UUID4, BaseModel
 
@@ -11,7 +11,7 @@ class InspectionError(BaseModel):
     Inspection error details
     """
 
-    document_id: PydanticObjectId | UUID4
+    document_id: Union[PydanticObjectId, UUID4]
     error: str
 
 
--- a/beanie/odm/models.py
+++ b/beanie/odm/models.py
@@ -1,6 +1,6 @@
 from typing import List
 
-from pydantic import BaseModel, UUID4
+from pydantic import UUID4, BaseModel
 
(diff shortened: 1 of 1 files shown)


beanie/odm/models.py
```diff
@@ -11,7 +11,7 @@ class InspectionError(BaseModel):
     Inspection error details
     """
 
-    document_id: PydanticObjectId
+    document_id: Union[PydanticObjectId, UUID4]
```
07pepa, line 14: since you can technicaly overide id (needs to be tested) with anything that is serializable by beanie... this is  is probably not enough (we should probably create Union of  Anything serializable probably something here https://github.com/BeanieODM/beanie/blob/54e0f2d5b2f3e9c83f37393fdd8bde4bf38fde81/beanie/odm/utils/encoder.py#L30 ) (somewhere close to Here
we have str as id and this would not work

I also do not see any tests  and it appears there is techical Debt in InspectionErrorCoverage


it is however coverable...  you insert with another library somethig that would triger InspectionError

second atitional test is to test if such type can be used as id (so use union of serializable to do that)

07pepa (changes requested):

PR #960, opened 2024-07-02
Feature / Fix: Allow Settings to be inherited and extended (fixes #644)

### LINKED ISSUE #644: [BUG] get_settings() does not return the inherited settings
**Describe the bug**

When inheriting settings from a common class, get_settings() does not return the inherited settings.

**To Reproduce**

```
from beanie import Document, init_beanie
import asyncio
import datetime


class GlobalSettings(Document):
    class Settings:
        bson_encoders = {
            datetime.datetime: lambda x: x,
            datetime.date: lambda x: datetime.datetime.combine(x, datetime.time.min),
        }


class MyDates(GlobalSettings):
    d: datetime.date
    dt: datetime.datetime

    class Settings(GlobalSettings.Settings):
        name = 'mydates'


async def example():
    client = AsyncIOMotorClient("mongodb://localhost:27019/my_database")
    await init_beanie(database=client.get_database(), document_models=[MyDates])
    doc = MyDates(d=datetime.date.today(), dt=datetime.datetime.now())
    print(doc.Settings.bson_encoders)
    print(doc.get_settings().bson_encoders)


if __name__ == "__main__":
    asyncio.run(example())
```

**Expected behavior**

doc.get_settings().bson_encoders (in this case) should return the same as doc.Settings.bson_encoder, i.e. the values from the inherited class.

**Additional context**
If
```
    class Settings(GlobalSettings.Settings):
        name = 'mydates'
```
 is removed, get_settings() works as expected.

dotKokott:
Before this would not work

```
from beanie import Document, init_beanie

class Sample1(Document):
    class Settings:
        name = "sample1"
        bson_encoders = {Color: lambda x: x.value}

class Sample2(Sample1):
    class Settings(Sample1.Settings):
        name = "sample2"
        
await init_beanie(
    database=db,
    document_models=[Sample2],
)        

assert Sample2.get_settings().bson_encoders != {} # Fails because Settings subclass completely overrides Sample1.Settings
```

The [fastapi-users](https://github.com/fastapi-users/fastapi-users) library actually claims in their documentation that this is how you should extend their `BeanieBaseUser` model to change the collection name: https://fastapi-users.github.io/fastapi-users/latest/configuration/databases/beanie/

Indeed the functionality would be very useful e.g. to provide shared `bscon_encoders` via a kind of base document model that all your documents inherit from.

This fixes https://github.com/BeanieODM/beanie/issues/644 which was closed without resolution. In the issue @roman-right mentions that this is not fixable because it would have to be fixed via `metaclass` which is already occupied by pydantic. Evidently, by this PR, that is not the case :)

The way the Settings attribute collection was implemented before (via `vars(class)`) would throw away the inherited `Settings` if a `Settings` class was present in the inheriting class.

I changed the `Settings` attribute collection to use `dir` and included some filtering of dunder methods.

✅ Tests passing  
✅ Added test to cover new functionality  

I don't know if this fix / feature needs a change in the [Settings](https://beanie-odm.dev/tutorial/defining-a-document/) documentation. Arguably it now behaves the way you'd expect this to behave, as it mirrors the way these attributes access and inherit via normal Python.


07pepa (approved):

adeelsohailahmed (approved):
Looks good to me. Thank you for your contribution.

Riverfount (approved):

PR #989, opened 2024-08-02
Fix regex storing

### LINKED ISSUE #983: No support for Regex elements
Hello,

In the process to migrate from odmantics to beanie, I need to process a List of Regex elements:

```
from odmantic import Model, Field
from odmantic.bson import Regex
class MyClass(Model):
    myField: Optional[List[Regex]] = None
```

I haven't figured out how to do it properly with beanie.
What I managed so far is to declare myField as List[Any] to read the Regex elements in my database and convert then to strings on the go when reading. But this solution can't store Regex, only strings.

```
def transform_to_str(input_data: List) -> List[Regex]:
    regex_list = []
    for element in input_data:
        if isinstance(element, str):
            regex_list.append(element)
        if isinstance(element, Regex):
            regex_list.append(element.pattern)
    return regex_list


class MyClass(Document):

    myField: Optional[List[Any]] = Field(default=None)

    @field_validator(
        'myField'
    )
    def convert_bson_regex(cls, value):
        return transform_to_str(value)
```

This solution seems to work but it feels a bit clunky.
Am i missing something ?

_Originally posted by @lothesven-clauger in https://github.com/BeanieODM/beanie/discussions/977_

07pepa:
Closes #983

### DIFF
--- a/beanie/odm/custom_types/re.py
+++ b/beanie/odm/custom_types/re.py
@@ -1,0 +1,26 @@
+import re
+
+import bson
+import pydantic
+from typing_extensions import Annotated
+
+from beanie.odm.utils.pydantic import IS_PYDANTIC_V2
+
+
+def _to_bson_regex(v):
+    return v.try_compile() if isinstance(v, bson.Regex) else v
+
+
+if IS_PYDANTIC_V2:
+    Pattern = Annotated[
+        re.Pattern,
+        pydantic.BeforeValidator(
+            lambda v: v.try_compile() if isinstance(v, bson.Regex) else v
+        ),
+    ]
+else:
+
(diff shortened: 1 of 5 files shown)


LawMixer (approved):

adeelsohailahmed (approved):

LawMixer (approved):

Riverfount (approved):

adeelsohailahmed (changes requested):
Please update the title of the PR and make it more descriptive.

07pepa:
@adeelsohailahmed  done

Riverfount (approved):

adeelsohailahmed (approved):
LGTM

PR #991, opened 2024-08-04
Add Operators: PushEach & AddToSetEach 

singh-lokendra:
Add Support for 

- pushEach: https://www.mongodb.com/docs/manual/reference/operator/update/each/
- addToSet: https://www.mongodb.com/docs/manual/reference/operator/update/addToSet/


adeelsohailahmed (changes requested):
Thank you for the PR.

Given that we already have the `AddToSet` and `Push`, wouldn't it be better to add just the `Each` operator and let user combine both operators explicitly instead?

PR #1015, opened 2024-09-04
Use sequence instead of list in init_beanie

07pepa:


Riverfount (approved):
LGTM

LawMixer (approved):
Looks good!

beanie/odm/utils/init.py
```diff
@@ -1,6 +1,8 @@
 import asyncio
 import sys
 
+from typing_extensions import Sequence, get_args, get_origin
+
```
adeelsohailahmed, line 5: I'm not sure if this is the best idea.

`typing_extensions` is supposedly only installed if Python version is less than 3.11, as per the pyproject.
image

Wouldn't this fail for those with Python version >= 3.11?

The only reason why the tests are not failing for Python >= 3.11 is because the typing_extensions is getting installed as part of fastapi (which is defined as an optional dependency for testing).

mg3146: What's the benefit of a sequence?

07pepa: @adeelsohailahmed now this is uniform for everything.... it will simplify import logic everywhere

@mg3146 just typing... anything that can be iterated over by for should be valid input

adeelsohailahmed (approved):
I've checked the `typing_extensions`.

It implicitly uses `get_origin` & `get_args` from `typing` module for Python versions 3.10 and above. So, indeed, this should work same as before. Good work!

Riverfount (approved):
That's ok to me.

07pepa:
> I've checked the `typing_extensions`.
> 
> It implicitly uses `get_origin` & `get_args` from `typing` module for Python versions 3.10 and above. So, indeed, this should work same as before. Good work!

yes typing extension are just basicly compatible backports for types of future versions almost as [pollyfill](https://en.wikipedia.org/wiki/Polyfill_(programming)) code for js

LawMixer (approved):

PR #1018, opened 2024-09-05
Use session in document insert

andraghetti:
# Summary
This PR fixes a transactional integrity issue in Beanie by ensuring that document saves are correctly associated with the current transaction session.

## Problem
When saving documents within a transaction using Beanie, a bug causes linked documents to be saved outside the transaction context if the session parameter is not explicitly provided. This behavior can lead to partial updates where some documents are persisted even if the transaction fails, undermining the purpose of using transactions.

## Solution
The change explicitly adds `session=session` to the `save` function for documents within a transaction. This ensures that all document saves are bound to the active transaction, preventing unintended persistence of documents if the transaction is rolled back.

This fix maintains the atomicity of transactions by ensuring all operations are executed within the transaction context, thus preserving data consistency and integrity.

### DIFF
--- a/beanie/odm/documents.py
+++ b/beanie/odm/documents.py
@@ -346,7 +346,9 @@ class Document(
                         LinkTypes.OPTIONAL_DIRECT,
                     ]:
                         if isinstance(value, Document):
-                            await value.save(link_rule=WriteRules.WRITE)
+                            await value.save(
+                                link_rule=WriteRules.WRITE, session=session
+                            )
                     if field_info.link_type in [
                         LinkTypes.LIST,
                         LinkTypes.OPTIONAL_LIST,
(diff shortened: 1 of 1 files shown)


adeelsohailahmed (approved):
Good catch! Thank you for your contribution. Looks good to me.

andraghetti:
@adeelsohailahmed Hi! Is there a way to expedite the process? Can we ping other collaborators to review?

adeelsohailahmed:
> @adeelsohailahmed Hi! Is there a way to expedite the process? Can we ping other collaborators to review?

Beyond requesting a review from the team (which I've already done), I don't think there's much we can do.

andraghetti:
> > @adeelsohailahmed Hi! Is there a way to expedite the process? Can we ping other collaborators to review?
> 
> Beyond requesting a review from the team (which I've already done), I don't think there's much we can do.

That’s perfect! Thank you 👍

ayusshsoarav (approved):

Riverfount (approved):
LGTM

andraghetti:
Nice! Thank you! Can somebody merge it?

Issue #1008, opened 2024-08-26
[BUG] BackLink causes error when using model_json_schema()


gdlmx:
**Describe the bug**
Generating Json Schema from a document with a `BackLink` field will fail with the error: `PydanticInvalidForJsonSchema`.

**To Reproduce**
```python
from typing import List
from pydantic import Field
from beanie import Document, Link, BackLink

class Task(Document):
    detail: str
    owner:  Link["User"]

class User(Document):
    name: str
    tasks: List[BackLink["Task"]] = Field(json_schema_extra={"original_field": "owner"})

User.model_json_schema()
```

The last line will raise error:

```python
PydanticInvalidForJsonSchema: Cannot generate a JsonSchema for core_schema.PlainValidatorFunctionSchema ({'type': 'with-info', 'function': <function BackLink.build_validation.<locals>.validate at 0x7f89140bdd00>})
```

**Expected behavior**
The last line (model_json_schema) should return 

```json
{
  "properties": {
    "_id": { "...":"..." },
    "name": {  "title": "Name",  "type": "string"  },
    "tasks": { "...":"..." }
  },
  "required": [ "name", "tasks" ],
  "title": "User",
  "type": "object"
}
```

**Additional context**
Python version: 3.11
Beanie version: 1.26.0
Pydantic version: 2.8.2

gdlmx:
Possible cause of this error:

https://github.com/BeanieODM/beanie/blob/bdb6a8692c87ff289c61e6b7df608e30767a94d8/beanie/odm/fields.py#L409-L430

See this [discussion](https://stackoverflow.com/a/77105412/11122845) of a similar error with Pydantic v2.4

adeelsohailahmed:
Thanks for reporting the issue, and taking the time to investigate the cause. Would you like to send a PR to fix this issue, @gdlmx?

xalien10:
@adeelsohailahmed @gdlmx Any updated about this fix yet?

gdlmx:
> Thanks for reporting the issue, and taking the time to investigate the cause. Would you like to send a PR to fix this issue, @gdlmx?

Yes, I can prepare a PR to fix it. Could you provide me some idea about the proper json schema for BackLink?

There is no guarantee of any field in a BackLink object, is there?

adeelsohailahmed:
> There is no guarantee of any field in a BackLink object, is there?

I'm not sure if I fully understand your question here. Could you elaborate more?

xalien10:
[...]

PR #1019, opened 2024-09-05
Fix bug #1008 in the json_schema export of document with BackLink

gdlmx:
This PR fixes bug #1008 and adds test for the model_json_schema() method of a document with a BackLink.


adeelsohailahmed (commented):
Thank you for your contribution. Several of the tests are failing. Could you please look into them?

gdlmx:
I have changed the title as requested. The failed test is caused by a [bug](https://github.com/pydantic/pydantic/issues/3390) in Pydantic V1. I cannot find an elegant solution, so I have to disable the `test_json_schema_export` for V1.

Riverfount (approved):
LGTM

PR #1022, opened 2024-09-06
fix type hint using pymongo client session instead of motor client session

CAPITAINMARVEL:
this pull request  fix the type hint of session

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,9 @@ pip install beanie
 ```shell
 poetry add beanie
 ```
+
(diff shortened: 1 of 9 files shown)


Riverfount (dismissed):
LGTM

07pepa (dismissed):

Riverfount (approved):
LGTM

Riverfount (approved):
I think it is ok!

LawMixer (approved):

staticxterm (approved):
Looks good. The only potential problem may be when syncing Bunnet with latest Beanie, but I'm not sure how that is done exactly. From the perspective of Beanie alone, this is not a breaking change. Thank you.

PR #1027, opened 2024-09-18
Set max pymongo version

kewldan:
Temporary solution:

A new version of pymongo has been released. Beanie doesn't support it yet. When installing motor, the new version 4.9 will be picked up. There will be an ImportError when running the application:

```
Traceback (most recent call last):
  File "/usr/app/src/main.py", line 1, in <module>
    from beanie import init_beanie
  File "/usr/local/lib/python3.12/site-packages/beanie/__init__.py", line 1, in <module>
    from beanie.migrations.controllers.free_fall import free_fall_migration
  File "/usr/local/lib/python3.12/site-packages/beanie/migrations/controllers/free_fall.py", line 4, in <module>
    from beanie.migrations.controllers.base import BaseMigrationController
  File "/usr/local/lib/python3.12/site-packages/beanie/migrations/controllers/base.py", line 4, in <module>
    from beanie.odm.documents import Document
  File "/usr/local/lib/python3.12/site-packages/beanie/odm/documents.py", line 69, in <module>
    from beanie.odm.interfaces.aggregate import AggregateInterface
  File "/usr/local/lib/python3.12/site-packages/beanie/odm/interfaces/aggregate.py", line 7, in <module>
    from beanie.odm.queries.aggregation import AggregationQuery
  File "/usr/local/lib/python3.12/site-packages/beanie/odm/queries/aggregation.py", line 12, in <module>
    from motor.core import AgnosticCommandCursor
  File "/usr/local/lib/python3.12/site-packages/motor/core.py", line 30, in <module>
    from pymongo.cursor import _QUERY_OPTIONS, Cursor, RawBatchCursor
ImportError: cannot import name '_QUERY_OPTIONS' from 'pymongo.cursor' (/usr/local/lib/python3.12/site-packages/pymongo/cursor.py)
```

What I did:
Wrote **pymongo<4.9** as a library dependency in pyproject.toml

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ classifiers = [
 dependencies = [
     "pydantic>=1.10,<3.0",
     "motor>=2.5.0,<4.0.0",
+    "pymongo<4.9",
     "click>=7",
     "toml",
     "lazy-model==0.2.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,8 +24,8 @@ classifiers = [
 ]
 dependencies = [
     "pydantic>=1.10,<3.0",
-    "motor>=2.5.0,<4.0.0",
-    "pymongo<4.9",
+    "motor~=3.6.0",
+    "pymongo~=4.9.1",
     "click>=7",
     "toml",
     "lazy-model==0.2.0",


adeelsohailahmed (dismissed):
Good catch! Thanks!

07pepa (approved):

07pepa (changes requested):
pymongo 4.9.1 seems to work can you try pymongo!=4.9 ?

kewldan:
No, motor 3.5.1 **doesnt** support 4.9.1. BUT motor has released a 3.6.0 which added a support for pymongo 4.9. So problem is solved. **Upgrade your motor library to 3.6.0**

adeelsohailahmed:
> No, motor 3.5.1 **doesnt** support 4.9.1. BUT motor has released a 3.6.0 which added a support for pymongo 4.9. So problem is solved. **Upgrade your motor library to 3.6.0**

Should we pin the motor to 3.6.0 then?

kewldan:
> > No, motor 3.5.1 **doesnt** support 4.9.1. BUT motor has released a 3.6.0 which added a support for pymongo 4.9. So problem is solved. **Upgrade your motor library to 3.6.0**
> 
> Should we pin the motor to 3.6.0 then? I'm fine with either/or.

Good question, it seems to me that yes, you should also specify pymongo~=4.9.1 since such mishaps happen. I would still mark all  current used versions as supported, so that a developer who uses beanie can be sure that these dependency versions will work correctly. I'll make the changes now

kewldan:
Interesting situation, motor latest version supports Python >= 3.8. Although support for Python 3.8 will end in 10 days (full, even security updates), so is it worth the effort and time to support it? This is more of a question for you here, since even the problem with pymongo is related to motor dependencies, not beanie dependencies

07pepa (approved):
for me i would drop anything that lost support in 3 monts at most but for now we should keep some compatability

adeelsohailahmed (commented):
Let's hold off merging this branch until we deprecate Python 3.7

kewldan:
Motor requires `pymongo>=4.9,<4.10` from [source code](https://github.com/mongodb/motor/blob/master/requirements.txt)

PR #1034, opened 2024-09-30
Add tests on all major mongo version

07pepa:


07pepa:
> Is there a reason why specifically 3.12.3 ?

yes 3.12.4+ will cause pydantic issue with misssing recursion guard and we would need to do biger rewrite of some testing infra see  https://github.com/BeanieODM/beanie/actions/runs/11106074398/job/30853640118

Riverfount (approved):
I think it is a great job.

LawMixer (approved):

PR #1037, opened 2024-10-05
Version/1.27.0 

adeelsohailahmed:


docs/changelog.md
```diff
@@ -2,6 +2,160 @@
...
+### Replace deprecated datetime.utcnow with datetime.now
+- Author - [adeelsohailahmed](https://github.com/adeelsohailahmed)
+- PR <https://github.com/BeanieODM/beanie/pull/1014>
+### Use sequence instead of list in init_beanie
+- Author - [07pepa](https://github.com/07pepa)
+- PR <https://github.com/BeanieODM/beanie/pull/1015>
+### Fix expression type hint not allowing some type https://github.com/beanieodm/beanie/issues/1020
+- Author - [CAPITAINMARVEL](https://github.com/CAPITAINMARVEL)
+- PR <https://github.com/BeanieODM/beanie/pull/1023>
+### Fix expression type hint not allowing some type https://github.com/beanieodm/beanie/issues/1020
```
staticxterm, line 27: Not sure what is going on here but some entries are being repeated multiple times.
By checking the other Releases and their notes there doesn't seem to be any duplicated entries.
Perhaps the `generate_changelog.py` script should be updated to do some kind of deduplication...?

staticxterm (changes requested):

docs/changelog.md
adeelsohailahmed: Oof, good catch, @staticxterm!

I've removed all the duplicated entries, and ordered the change log based on the PR number for easier review. Could you please have another look?

staticxterm (approved):

roman-right (approved):
lgtm

docs/changelog.md
```diff
@@ -2,6 +2,97 @@
 
 Beanie project
 
+## [1.27.0] - 2024-10-06
```
staticxterm, line 5: Only one slight "issue" now with the approval coming late is that yesterdays date is for the 1.27.0 "release".
Also, we should also release the documentation, are there efforts to fix the docs publish action?

adeelsohailahmed: If we were to update the date, it would automatically dismiss the previous approvals, and require everyone to approve again. I think it's best if we just get this release out of the way and move on to the next release.

staticxterm: Agree. Just wanted to point this out.

Issue #1010, opened 2024-08-29
[BUG] FindMany count ignores limit set


pschoen-itsc:
**Describe the bug**
When getting to count of a query with a limit applied before, the limit gets ignored. That is a problem, because so mongodb has to find all. documents matching the query to count them, instead of stopping after reaching the limit. In big collections (big documents, millions of them) this has a big performance impact.

**To Reproduce**
```python
count = await MyDocument.find_many(MyDocument.foo == "bar").limit(1000).count()
```
does not return the count of all Documents with foo==bar, instead of maximum 1000.

**Expected behavior**
Returns 1000 if more documents than 1000 or found, otherwise the count lower than 1000.

**Additional context**
We moved from pure pymongo queries to beanie. With pymongo this worked as I described my expectation here. So it would be nice to have the same behaviour here.

Issue #848, opened 2024-01-31
[BUG] FindQuery count() function does not use session parameter


QSHolzner:
**Describe the bug**
The count() function of FindQuery does not use a session and therefore runs outside of a specified transaction context. In the implementation, the function count_documents() of AsyncIOMotorCollection is called, but the session parameter is not set there. 

**Current implemenation**
```python
 async def count(self) -> int:
        """
        Number of found documents
        :return: int
        """
        return (
            await self.document_model.get_motor_collection().count_documents(
                self.get_filter_query()
            )
        )
```

**Expected behavior**
If find() is called with a session parameter, this session should be used in the count() call. 

**Changed implementation**
This should fix the behaviour.
```python
async def count(self) -> int:
        """
        Number of found documents
        :return: int
        """
        return (
            await self.document_model.get_motor_collection().count_documents(
                self.get_filter_query(), session=self.session 
            )
        )
```

roman-right:
Good catch! Thank you

PR #1040, opened 2024-10-07
Handle limit and session in .count() method

CAPITAINMARVEL:
This PR is to fix the issue related to the count method

- Fixes https://github.com/BeanieODM/beanie/issues/1010
- Fixes https://github.com/BeanieODM/beanie/issues/848

### DIFF
--- a/tests/odm/documents/test_count.py
+++ b/tests/odm/documents/test_count.py
@@ -21,4 +21,3 @@ async def test_count_with_limit(documents):
     assert c == 1
     d = await DocumentTestModel.find_all().count()
     assert d == 5
-


LawMixer (approved):
simple pr, looks good!

adeelsohailahmed:
Closes #1010 & #848

adeelsohailahmed (approved):
LGTM

07pepa (approved):

07pepa:
this is marvelous work... pun intended

martincpt:
I just realised that this PR broke our CRUD app.

I don't think this functionality is quite right as it is, because it's breaking any implementation - especially pagination - that relies on the real count number.

Don't you think it should be optional instead?

PR #1042, opened 2024-10-08
Add DocumentDB compatibility to fetch_links

whitfin:
DocumentDB does not support `$set` and `$unset` in aggregation pipelines. This PR will change the pipelines for `fetch_links=True` to not use them.

The `$set` operator is synonymous with `$addFields` (no idea why), and `$unset` is an alias for `$project` setting the values to `0`. As such this is no more/less efficient, it's just an improvement in compatibility.

This is retarget of https://github.com/BeanieODM/bunnet/pull/19 which was closed due to inactivity, but I figure patching here will eventually get merged back over into Bunnet regardless.

FWIW this does not necessarily fix all compatibility; just this specific aspect.

### DIFF
--- a/beanie/odm/utils/find.py
+++ b/beanie/odm/utils/find.py
@@ -66,7 +66,7 @@ def construct_query(
                     }
                 },
                 {
-                    "$set": {
+                    "$addFields": {
                         link_info.field_name: {
                             "$cond": {
                                 "if": {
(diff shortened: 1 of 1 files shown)


07pepa (approved):

staticxterm (approved):
Thank you for the PR! Nice to see this as a step closer to supporting DocumentDB as well :)

adeelsohailahmed (approved):

Vert53:
Hey @whitfin. I'm getting this error when using fetch_links. Do you maybe have an idea if there is a way to get around this issue?
`pymongo.errors.OperationFailure: $lookup concise correlated subquery is not supported, full error: {'ok': 0.0, 'code': 9, 'errmsg': '$lookup concise correlated subquery is not supported', 'operationTime': Timestamp(1739107109, 1)}`

PR #1044, opened 2024-10-08
Drop support for Python 3.7

07pepa:
* modify gh actions
* improve dependency definitions

### DIFF
--- a/.github/workflows/github-actions-tests.yml
+++ b/.github/workflows/github-actions-tests.yml
@@ -6,20 +6,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
-          python-version: 3.11
+          python-version: 3.12
       - uses: pre-commit/action@v3.0.1
   run-tests:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.7", "3.8", "3.9", "3.10", "3.11", "3.12.3" ]
-        mongodb-version: [ 4.4, 5.0, 6.0, 7.0, 8.0 ]
-        pydantic-version: [ "1.10.15", "2.7" ]
-        exclude:
-          - python-version: "3.7"
-            pydantic-version: "2.7"
+        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12" ]
+        mongodb-version: [4.4, 5.0, 6.0, 7.0, 8.0 ]
+        pydantic-version: [ "1.10.18", "2.9.2" ]
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/github-actions-tests.yml
```diff
@@ -6,20 +6,17 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.7", "3.8", "3.9", "3.10", "3.11", "3.12.3" ]
-        mongodb-version: [ 4.4, 5.0, 6.0, 7.0, 8.0 ]
-        pydantic-version: [ "1.10.15", "2.7" ]
-        exclude:
-          - python-version: "3.7"
-            pydantic-version: "2.7"
+        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12"]
```
staticxterm, line 17: It probably doesn't matter one bit, but there is a missing space before the closing square bracket: _]_.

staticxterm (commented):
Hi @07pepa, the PR is good. However, I believe that it may be better if we split it into multiple ones. Either that or somehow change the PR title to be more succinct and just generally rephrased.

The main point as I understand it is the deprecation of Python 3.7 and the changes to the CI machinery and in pyproject.toml. Then we have the uplifts of development dependencies, some codebase modernization due to deprecations and a single Pydantic v2 Config compatibility fix...

I feel like the intent may better be conveyed if we have at least 3 separate PRs for this. But this is just my opinion, curious what others think.

07pepa:
> Hi @07pepa, the PR is good. However, I believe that it may be better if we split it into multiple ones. Either that or somehow change the PR title to be more succinct and just generally rephrased.
> 
> The main point as I understand it is the deprecation of Python 3.7 and the changes to the CI machinery and in pyproject.toml. Then we have the uplifts of development dependencies, some codebase modernization due to deprecations and a single Pydantic v2 Config compatibility fix in the test models...
> 
> I feel like the intent may better be conveyed if we have at least 3 separate PRs for this. But this is just my opinion, curious what others think.

yea this is kindy my bad habit of mine ....

keeping only pyproject changes and test changes

staticxterm:
Looks good, I mean there's again more than one change but that's fine I guess since it only impacts our development experience.
One last proposal though is to perhaps rename the PR title to 'Drop support for Python 3.7' or something along those lines since 'remove python' sounds a bit weird 😅
Next on our to do list of sorts should also be to see for how long we would want to support Py3.8 since it lost official support now as well...

adeelsohailahmed:
> Next on our to do list of sorts should also be to see for how long we would want to support Py3.8 since it lost official support now as well...

I suggest we support 3.8 for a month or so, at least.

LawMixer (approved):

staticxterm (approved):

PR #1046, opened 2024-10-09
update pre-commit

07pepa:
simple pr updating pre-commit


LawMixer (approved):
simple enough

07pepa:
> beanie/odm/queries/find.py:1033:12: E721 Use `is` and `is not` for type comparisons, or `isinstance()` for isinstance checks
> 
> That would be in another PR I assume ?

it shoul be here but was missed

LawMixer (approved):

PR #1047, opened 2024-10-09
modify tests to not raise deprecation warnings

07pepa:


beanie/odm/documents.py
```diff
@@ -1136,7 +1136,7 @@ async def inspect_collection(
         return inspection_result
 
     @classmethod
-    def check_hidden_fields(cls):
```
07pepa, line 1139: imho this should not be public method

tests/odm/conftest.py
```diff
@@ -197,118 +297,38 @@ async def session(cli):
     await s.end_session()
 
 
+@pytest.fixture()
+async def deprecated_init_beanie(db):
```
07pepa, line 301: i wanted to explicitly separate deprecated usages from not deprecated ones
i also added assertions those deprecations are actualy raised

LawMixer (approved):
the other review covered what i would change

tests/odm/conftest.py
```diff
@@ -197,118 +297,38 @@ async def session(cli):
...
+    for model in TESTING_MODELS:  # crude clear from init
+        await model.get_motor_collection().drop()
+        await model.get_motor_collection().drop_indexes()
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
+
+        await init_beanie(
```
staticxterm, line 308: Only one slight "issue" is that now `init_beanie` will be executed twice (but I guess this is fine), since there is an autouse `init` fixture in this file as well.

staticxterm (approved):

tests/odm/conftest.py
07pepa: yes geting rid of auto use would be hard.... i am fully aware of it...

PR #1049, opened 2024-10-10
Fix kwargs/args untyped

CAPITAINMARVEL:


staticxterm (commented):
Looks good. Please remove the stray session typing fix in beanie/odm/interfaces/aggregation_methods.py and leave it for some other commit, let's handle only kwargs/args typings in this PR.

Missed one in beanie/executors/migrate.py:
class MigrationSettings:
    def __init__(self, **kwargs):
    
But migrate.py in itself is very untyped and should probably be treated/fully typed out in some follow-up task (since we said that this package is fully typed by using the `py.typed`).

Other usages are in the `__call__` method in beanie/migrations/controllers/free_fall.py and in beanie/migrations/controllers/iterative.py. Then `__init__` methods in beanie/odm/documents.py and beanie/odm/operators/find/array.py, 
and lastly, in utils: beanie/odm/utils/pydantic.py and beanie/odm/utils/state.py.

This is all that a VSCode search spewed out for me. This should squash them all, then we move to the next ones :)

Riverfount (dismissed):
LGTM

beanie/odm/documents.py
```diff
@@ -691,13 +691,13 @@ async def replace_many(
     @save_state_after
     async def update(
         self: Self,
-        *args,
+        *args: Any,
```
staticxterm, line 694: Not aligned with the docstring below:
`:param args: *Union[dict, Mapping] - the modifications to apply.`
And docstring is generally out of date as it doesn't mention all new params.

staticxterm (approved):

LawMixer (approved):

LawMixer (approved):
LGTM!

Riverfount (approved):
LGTM, again! ;-)

staticxterm (approved):

07pepa (approved):

PR #1051, opened 2024-10-11
add python 3.13 and JIT into testing

07pepa:
* add python 3.13
* turn on JIT
* update code to not raise deprecation warnings

### DIFF
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -1,5 +1,6 @@
+from __future__ import annotations
+
 import asyncio
-import sys
 from collections import OrderedDict
 from dataclasses import dataclass
 from enum import Enum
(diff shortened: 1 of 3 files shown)


Riverfount (approved):
Great job.

beanie/odm/fields.py
```diff
@@ -1,5 +1,6 @@
+from __future__ import annotations
```
staticxterm, line 1: I remember there was some issue a while back with using this import and with lets say Pydantic.
This basically enables all new typing features and whatnot.

I'm fine with this addition if it doesn't break any downstream consumers (existing codebases).

On another note, does this mean that we can now start modernizing the Beanie codebase using modern typing features, such as `list[str] | None` instead of `Optional[List[str]]` and so on (by adding this import to every file)?

pyproject.toml
```diff
@@ -84,6 +84,7 @@ testpaths = [
 ]
 filterwarnings = [
     "error",
+    "ignore::DeprecationWarning",
```
staticxterm, line 87: In the PR description you write 
> update code to not raise deprecation warnings

since the code was updated to not use the "older typing syntax", can we please remove this suppression/ignore rule? I ran pytest locally and get no warnings with this line removed.

Riverfount (dismissed):
Two comments need to be answered before this PR can be approved.

beanie/odm/fields.py
07pepa: it should be fine for python 3.8+

pyproject.toml
07pepa: this is pydantic 1.x.x causing this .... see https://github.com/BeanieODM/beanie/actions/runs/11526533446/job/32090846596?pr=1062

and if we remove it then we can remove it

staticxterm: Okay @07pepa, my mistake then. Good point.

staticxterm (approved):
Thanks for answering the comments. Looks good.

Riverfount (approved):
LGTM

PR #1053, opened 2024-10-11
fix issues caused by #1044

07pepa:


staticxterm (approved):

LawMixer (approved):
i tried this version and had no issues - we should be fine

PR #1054, opened 2024-10-14
Fix `PydanticObjectId` fields being parsed into `str` when using Pydantic's `model_validate_json`

### LINKED ISSUE #940: [BUG] `PydanticObjectId` isn't properly configured when deserializing from JSON
**Describe the bug**
Consider the following code:
```python
class A(BaseModel):
   id: PydanticObjectId

raw = '{"id": "5eb7cf5a86d9755df3a6c593"}'
deserialized = A.model_validate_json(raw)
```

One would expect `type(deserialized.id) == PydanticObjectId`, but it's actually deserialized to `str`.

**To Reproduce**
See example above.

**Expected behavior**
`type(deserialized.id)` should be `PydanticObjectId`.

**Additional context**
Note that if I do `A.model_validate(json.loads(raw))` the code works. This is potentially because in the [`__get_pydantic_core_schema__` override in `PydanticObjectId`](https://github.com/BeanieODM/beanie/blob/main/beanie/odm/fields.py#L146), `python_schema` includes a transformation of the id from `str` to `ObjectId`, while the `json_schema` doesn't and directly uses the built-in `str_schema`.

ivan-gj:
Unsure if this could be fixed in a more stylish fashion, but it was simply incorrectly setup before, at least now it works. 

**Please make as many changes as needed if this solution could be improved.**

I believe this bugfix is quite important. If you parse an ID (let's call the variable `identifier`) from a JSON file in order to look it up on your Mongo database... you won't find it.

```python
 await cls.find_one(cls.id == identifier)
```
will return `None` even if the object exists, because `identifier` will be of type `str` instead of `ObjectId` when parsing from a JSON file with Pydantic's `model_validate_json`. 

Fixes #940.

### DIFF
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -159,7 +159,7 @@ class PydanticObjectId(ObjectId):
         ) -> CoreSchema:  # type: ignore
             return core_schema.json_or_python_schema(
                 python_schema=plain_validator(cls.validate),
-                json_schema=str_schema(),
+                json_schema=plain_validator(cls.validate),
                 serialization=core_schema.plain_serializer_function_ser_schema(
                     lambda instance: str(instance), when_used="json"
                 ),
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -58,7 +58,6 @@ if IS_PYDANTIC_V2:
     from pydantic_core.core_schema import (
         ValidationInfo,
         simple_ser_schema,
-        str_schema,
     )
 else:
     from pydantic.fields import ModelField  # type: ignore


staticxterm:
Hi @ivan-gj, thank you for your contribution.
So this only happens with Pydantic v2, right? Could you please add an example test case that was failing before, and is passing now (to cover this regression)? E.g. like the code from #940?

07pepa:
@ivan-gj  Thanks for your contribution. There is some undocumented pydantic stuff that i had to figure out with brute force in my pr  #1060  I wanted to help but ended up doing whole PR

closing as duplicate

ivan-gj:
No problem! Glad to see it getting fixed.

PR #1063, opened 2024-10-28
start to use Perflint to improved performance

07pepa:
added option to ruff to guide developers to use more performant constructs in python

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,12 @@
 repos:
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.6.9
+    rev: v0.7.1
     hooks:
       - id: ruff
         args: [ --fix ]
       - id: ruff-format
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.11.2
+    rev: v1.13.0
     hooks:
       - id: mypy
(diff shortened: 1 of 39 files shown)


kewldan (commented):
TL;DR: Partially useful changes

After looking at all your changes, I conclude that they don't affect performance much unless you're chasing nanoseconds. In general, of course, some changes should really be made to the code base. And some that might reduce execution time by n amount of time are not worth it, as they **make the code heavier to read**. I would consider to commit partial changes. 

IMHO

kewldan:
> > TL;DR: Partially useful changes
> > After looking at all your changes, I conclude that they don't affect performance much unless you're chasing nanoseconds. In general, of course, some changes should really be made to the code base. And some that might reduce execution time by n amount of time are not worth it, as they **make the code heavier to read**. I would consider to commit partial changes.
> > IMHO
> 
> like if else expression ? making it more difficult to read ?

Not only statements, complicated list/dict comprehension. I didn't say that all changes are useless or over complicated due to reduce execution time, I fully agree with you that some solutions in the current code have some flaws that you want to correct. Don't take it as disrespect 😁

07pepa:
python etos says readability counts and so just close this ...but i will get of those pesky gathers with nicer readable sintax from anyio

PR #1065, opened 2024-11-07
start using any io for structured concurency

07pepa:
Rationale

gather does not cancel other running tasks.

using structured concurency (aka TaskGroup) will cancel running if any of  corutines in same group fails.
This avoid unnecesary and unexpected memory leak performance because we will be running "zombie" tasks we no longer need throwing out results.

blurb from python documentation 
image
this is awailible in [python 3.12](https://docs.python.org/3/library/asyncio-task.html#asyncio.TaskGroup) and equivalent in [trio](https://trio.readthedocs.io/en/stable/reference-core.html#nurseries-and-spawning)

### DIFF
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import asyncio
 from collections import OrderedDict
 from dataclasses import dataclass
 from enum import Enum
@@ -18,6 +17,7 @@ from typing import (
 )
 from typing import OrderedDict as OrderedDictType
 
+from anyio import create_task_group
 from bson import DBRef, ObjectId
 from bson.errors import InvalidId
 from pydantic import BaseModel
@@ -363,10 +363,10 @@ class Link(Generic[T]):
 
     @classmethod
     async def fetch_many(cls, links: List[Link]):
-        coros = []
-        for link in links:
-            coros.append(link.fetch())
-        return await asyncio.gather(*coros)
+        if links:
+            async with create_task_group() as tg:
+                for link in links:
+                    tg.start_soon(link.fetch)
 
     if IS_PYDANTIC_V2:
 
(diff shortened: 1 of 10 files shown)


Riverfount (approved):
LGTM

beanie/odm/fields.py
```diff
@@ -363,10 +363,10 @@ def repack_links(
 
     @classmethod
     async def fetch_many(cls, links: List[Link]):
-        coros = []
-        for link in links:
-            coros.append(link.fetch())
-        return await asyncio.gather(*coros)
```
staticxterm, line 369: This method used to return here previously. I don't see how are the "results" being returned now?

beanie/odm/fields.py
```diff
@@ -363,10 +363,10 @@ def repack_links(
...
+        if links:
+            async with create_task_group() as tg:
+                for link in links:
+                    tg.start_soon(link.fetch)
```
staticxterm, line 369: If I'm not mistaken (since I don't know anyio), here we should use something else apart from tg.start_soon() since it can't be awaited...
Alternatively, if somehow these results can be collected afterwards then we await all those tasks and return all the collected results.

staticxterm (changes requested):
Looks good generally. Found one small breaking change.
Personally, since this is kind of a bigger change, I have fears if something else may break due to this, since we have no tests to verify if everything is working as it should. So I wouldn't feel very confident using this in a production app.

beanie/odm/fields.py
staticxterm: E.g. FastAPI does something like this, lines 856-867:
https://github.com/fastapi/fastapi/blob/bffb4115a9b63127948cc5e1aa14d73940734f75/fastapi/dependencies/utils.py#L856
It calls tg.start_soon(), but then in the method it calls it stores those results to a variable declared in the "upper" scope. Not sure if this is feasible to do here, since we call some "internal" Link class method. Perhaps feasible, but with some more refactoring required...

beanie/odm/fields.py
07pepa: good catch. will add test to his this is breaking change

roman-right:
@roman-right to take a look

PR #1071, opened 2024-11-23
Fix AnyUrl not being encodable

07pepa:
On nevest pydantic  anyurl is not ecodable

### DIFF
--- a/.github/workflows/github-actions-tests.yml
+++ b/.github/workflows/github-actions-tests.yml
@@ -16,7 +16,7 @@ jobs:
       matrix:
         python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12", "3.13" ]
         mongodb-version: [4.4, 5.0, 6.0, 7.0, 8.0 ]
-        pydantic-version: [ "1.10.18", "2.9.2" ]
+        pydantic-version: [ "1.10.18", "2.10.1" ]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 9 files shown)


staticxterm:
Looks good, please add some tests to tests/odm/test_encoder.py or elsewhere?
How come that the current test case "test_should_encode_pydantic_v2_url_correctly" didn't catch this?

07pepa:
> Looks good, please add some tests to tests/odm/test_encoder.py or elsewhere? How come that the current test case "test_should_encode_pydantic_v2_url_correctly" didn't catch this?

there already is test that would fail without change ... hiden by using older version upgrading version uncovered this.

staticxterm:
> > Looks good, please add some tests to tests/odm/test_encoder.py or elsewhere? How come that the current test case "test_should_encode_pydantic_v2_url_correctly" didn't catch this?
> 
> there already is test that would fail without change ... hiden by using older version upgrading version uncovered this.

Okay, I get this, but I believe there should be a kind of a test that would catch this regression, if possible. But I see you've added the OpenAPI schema generation test. I guess this could suffice for now.
By the way, did this test fail under Pydantic v1 only, under Pydantic v2 only, under some specific Pydantic version or what was the case here?

staticxterm (changes requested):

Riverfount (changes requested):
Just review the notes of @staticxterm and the Github Action is failing, fix this issues.

staticxterm:
Looked into this a bit, it seems there's a lot of breakage with Url types in Pydantic v2.10.x. It was still broken for me on current latest 2.10.3. I suggest to reduce this PR to a minimum change in project.toml where we skip the Pydantic minor version range which is causing these issues (works fine on 2.9.2, started breaking with 2.10.x) and lock it down to a known working version (2.9.2) in our CI as well. Hopefully Pydantic team fixes it upstream with 2.11.0+ or some other patch version on 2.10.x (doesn't seem like it).

PR #1079, opened 2024-12-06
Bulk Writer improving & bulk_writer method for document and possibility to bypass mongo document validation + comment parameter

CAPITAINMARVEL:
- Reworked it
added:
- bypass_document_validation
- comment parameter
- method to use bulk_write on Document and UnionDoc model

handle
https://github.com/BeanieODM/beanie/discussions/327

### DIFF
--- a/beanie/odm/bulk.py
+++ b/beanie/odm/bulk.py
@@ -44,7 +44,7 @@ class BulkWriter:
             Specifies whether operations are executed sequentially (default) or in parallel.
             - If True, operations are performed serially, stopping at the first failure.
             - If False, operations may be executed in arbitrary order, and all operations are attempted
-              regardless of individual failures.
+            regardless of individual failures.
         bypass_document_validation Optional[bool]:
             If True, document-level validation is bypassed for all operations
             in the bulk write. This applies to MongoDB's schema validation rules, allowing documents that
@@ -54,6 +54,8 @@ class BulkWriter:
             auditing and debugging purposes.
         operations List[Union[DeleteMany, DeleteOne, InsertOne, ReplaceOne, UpdateMany, UpdateOne]]:
             A list of MongoDB operations queued for bulk execution.
+        object_class Type[Union[Document, UnionDoc]]:
+            The document model class associated with the operations.
 
     Parameters:
         session Optional[AsyncIOMotorClientSession]: The motor session for transaction support.
@@ -65,19 +67,21 @@ class BulkWriter:
             where strict validation may not be necessary. Defaults to False.
         comment Optional[Any]: A custom comment attached to the bulk operation.
             Defaults to None.
+        object_class Type[Union[Document, UnionDoc]]: The document model class associated with the operations.
     """
 
     def __init__(
         self,
         session: Optional[AsyncIOMotorClientSession] = None,
         ordered: bool = True,
+        object_class: Optional[Type[Union[Document, UnionDoc]]] = None,
         bypass_document_validation: bool = False,
         comment: Optional[Any] = None,
     ) -> None:
         self.operations: List[_WriteOp] = []
         self.session = session
         self.ordered = ordered
-        self.object_class: Optional[Type[Union[Document, UnionDoc]]] = None
+        self.object_class = object_class
         self.bypass_document_validation = bypass_document_validation
         self.comment = comment
         self._collection_name: str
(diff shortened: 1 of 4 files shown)


tests/odm/documents/test_bulk_write.py
```diff
@@ -182,3 +182,31 @@ async def test_native_upsert_not_found(documents, document_not_inserted):
...
+
+
+async def test_comment(documents_not_inserted):
+    bulk = BulkWriter(comment="test")
+    assert bulk.comment == "test"
+    bulk2 = BulkWriter()
+    assert not bulk2.comment
+
+
+async def test_bypass_document_validation(documents_not_inserted):
```
staticxterm, line 208:
```suggestion
async def test_bypass_document_validation():
```
Same as previous comment, the `documents_not_inserted` fixture is not needed.
Test is passing even without it.

tests/odm/documents/test_bulk_write.py
```diff
@@ -182,3 +182,31 @@ async def test_native_upsert_not_found(documents, document_not_inserted):
...
+    except ValueError:
+        pytest.fail("ValueError raised unexpectedly")
+
+
+async def test_empty_operations(documents_not_inserted):
+    bulk = BulkWriter()
+    assert await bulk.commit() == None
+
+
+async def test_comment(documents_not_inserted):
```
staticxterm, line 201:
```suggestion
async def test_comment():
```
Nothing is being inserted to the DB, so the `documents_not_inserted` fixture is not needed.
Test is passing even without it.

staticxterm (changes requested):
The BulkWriter changes look good, needs some more polishing...

beanie/odm/bulk.py
```diff
@@ -12,82 +13,135 @@
...
+            Defaults to None.
+    """
+
     def __init__(
         self,
         session: Optional[AsyncIOMotorClientSession] = None,
         ordered: bool = True,
+        bypass_document_validation: bool = False,
+        comment: Optional[Any] = None,
     ):
```
staticxterm, line 75:
```suggestion
    ) -> None:
```
In an effort to fully type annotate everything, mypy under strict mode requires "-> None" typing in such cases.

beanie/odm/bulk.py
```diff
@@ -12,82 +13,135 @@
...
-        self.operations: List[Operation] = []
+        self.operations: List[_WriteOp] = []
         self.session = session
         self.ordered = ordered
+        self.object_class: Optional[Type[Union[Document, UnionDoc]]] = None
+        self.bypass_document_validation = bypass_document_validation
+        self.comment = comment
+        self._collection_name: str
 
     async def __aenter__(self):
```
staticxterm, line 84:
```suggestion
    async def __aenter__(self) -> "BulkWriter":
```

beanie/odm/bulk.py
```diff
@@ -12,82 +13,135 @@
...
         return self
 
     async def __aexit__(self, exc_type, exc, tb):
```
staticxterm, line 87:
```suggestion
    async def __aexit__(
        self,
        exc_type: Optional[Type[BaseException]],
        exc: Optional[BaseException],
        tb: Optional[TracebackType],
    ) -> None:
```
```python
from types import TracebackType
```

staticxterm (changes requested):
Looks good, have one final nitpick about one test assertion and some typing improvements if you agree...

staticxterm (approved):
Looks great, thank you!

adeelsohailahmed (approved):
Good stuff! Please update the PR's title to be a bit more descriptive for the change log.

beanie/odm/documents.py
```diff
@@ -1214,6 +1213,43 @@ def link_from_id(cls, id: Any):
         ref = DBRef(id=id, collection=cls.get_collection_name())
         return Link(ref, document_class=cls)
 
+    def bulk_writer(
```
roman-right, line 1216: Could you please add a test for this method? In case something will be changed later it will prevent breaking changes

beanie/odm/union_doc.py
```diff
@@ -38,3 +41,40 @@ def register_doc(cls, name: str, doc_model: Type):
     @classmethod
     def get_model_type(cls) -> ModelType:
         return ModelType.UnionDoc
+
+    def bulk_writer(
```
roman-right, line 45: Please, add a test for this one as well

beanie/odm/bulk.py
```diff
@@ -12,82 +14,140 @@
...
         ordered: bool = True,
-    ):
-        self.operations: List[Operation] = []
+        bypass_document_validation: bool = False,
+        comment: Optional[Any] = None,
+    ) -> None:
+        self.operations: List[_WriteOp] = []
         self.session = session
         self.ordered = ordered
+        self.object_class: Optional[Type[Union[Document, UnionDoc]]] = None
```
roman-right, line 80: I think you can add object class as a constructor parameter here. It should not break any logic but will allow set up the correct class on the very first step

roman-right (changes requested):
Please take a look at the few comments I left.
In general, it is a great job! Thank you!

roman-right (approved):

adeelsohailahmed (approved):

staticxterm (approved):

PR #1089, opened 2024-12-12
Make `diacritic_sensitive` parameter optional to support $text operator on Cosmos DB

### LINKED ISSUE #1088: [BUG] Text search in Azure CosmosDB fails due to the lack of diacritic sensitivity support
**Describe the bug**
It is not possible to perform a full-text search on Azure Cosmos DB for MongoDB due to the fact that [diacritic sensitivity is not supported](https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/vcore/how-to-create-text-index) yet.

**To Reproduce**
```python
import asyncio
import os

from beanie import Document, Indexed, init_beanie
from beanie.operators import Text
from pymongo import TEXT


class Dish(Document):
    name: Indexed(str, index_type=TEXT)


async def example():
    client = AsyncIOMotorClient(os.environ["COSMOSDB_CONNECTION"])
    await init_beanie(database=client["test"], document_models=[Dish])
    names = [
        "Ravioli",
        "Spaghetti al pomodoro",
        "Tagliatelle al pomodoro",
    ]
    for name in names:
        dish = Dish(name=name)
        await dish.insert()

    dish = await Dish.find_one(Dish.name == "Ravioli")
    print("dish", dish)
    # dish id=ObjectId('675b131c7b21f749a28f32ff') revision_id=None name='Ravioli'
    result = await Dish.find(Text("pomodoro")).to_list()
    # pymongo.errors.OperationFailure: $diacritic insensitive searches are not supported yet
    print("result", result)


if __name__ == "__main__":
    asyncio.run(example())
```

**Expected behavior**
The code should have produced the search results for a text query, but since the [`$diacriticSensitive` is always present](https://github.com/mykolaskrynnyk/beanie/blob/0773aed151acb30dfcc4448cdbc26326c7366a53/beanie/odm/operators/find/evaluation.py#L165-L176) in the command, the search fails with

```python
pymongo.errors.OperationFailure: $diacritic insensitive searches are not supported yet, full error: {'ok': 0.0, 'errmsg': '$diacritic insensitive searches are not supported yet', 'code': 115, 'codeName': 'CommandNotSupported'}
```
**Additional context**
I will patch it up shortly.

mykolaskrynnyk:
Closes #1088.

### DIFF
--- a/beanie/odm/operators/find/evaluation.py
+++ b/beanie/odm/operators/find/evaluation.py
@@ -148,14 +148,14 @@ class Text(BaseFindEvaluationOperator):
         search: str,
         language: Optional[str] = None,
         case_sensitive: bool = False,
-        diacritic_sensitive: bool = False,
+        diacritic_sensitive: Optional[bool] = False,
     ):
         """
 
         :param search: str
         :param language: Optional[str] = None
         :param case_sensitive: bool = False
-        :param diacritic_sensitive: bool = False
+        :param diacritic_sensitive: Optional[bool] = False
         """
         self.search = search
         self.language = language
@@ -168,11 +168,14 @@ class Text(BaseFindEvaluationOperator):
             "$text": {
                 "$search": self.search,
                 "$caseSensitive": self.case_sensitive,
-                "$diacriticSensitive": self.diacritic_sensitive,
             }
         }
         if self.language:
             expression["$text"]["$language"] = self.language
+        if self.diacritic_sensitive is not None:
+            expression["$text"]["$diacriticSensitive"] = (
+                self.diacritic_sensitive
+            )
         return expression
 
 
--- a/beanie/odm/operators/find/evaluation.py
+++ b/beanie/odm/operators/find/evaluation.py
@@ -141,6 +141,10 @@ class Text(BaseFindEvaluationOperator):
 
     MongoDB doc:
     <https://docs.mongodb.com/manual/reference/operator/query/text/>
+
+    Note: if you need to run a query against Azure Cosmos DB for MongoDB,
+    which does not support diacritic sensitivity yet, you can set
+    `diacritic_sensitive` argument to `None` to exclude it from the query.
     """
 
     def __init__(
--- a/tests/odm/operators/find/test_evaluation.py
+++ b/tests/odm/operators/find/test_evaluation.py
@@ -57,6 +57,13 @@ async def test_text():
             "$diacriticSensitive": True,
         }
     }
+    q = Text("something", diacritic_sensitive=None)
+    assert q == {
+        "$text": {
+            "$search": "something",
+            "$caseSensitive": False,
+        }
+    }
     q = Text("something", language="test")
     assert q == {
         "$text": {


mykolaskrynnyk:
To use text search with Azure CosmosDB for MongoDB, set `diacritic_sensitive` to `None`, e.g.,


```python
import asyncio
import os

from beanie import Document, Indexed, init_beanie
from beanie.operators import Text
from pymongo import TEXT


class Dish(Document):
    name: Indexed(str, index_type=TEXT)


async def example():
    client = AsyncIOMotorClient(os.environ["COSMOSDB_CONNECTION"])
    await init_beanie(database=client["test"], document_models=[Dish])
    names = [
        "Ravioli",
        "Spaghetti al pomodoro",
        "Tagliatelle al pomodoro",
    ]
    for name in names:
        dish = Dish(name=name)
        await dish.insert()
    result = await Dish.find(Text("pomodoro", diacritic_sensitive=None)).to_list()
    print("result", result)


if __name__ == "__main__":
    asyncio.run(example())
```

adeelsohailahmed (changes requested):
Hi @mykolaskrynnyk, thank you for your contribution! 

Please update the docstring to explain why `diacritic_sensitive` would be set to `None`.

Please update this test to address the case where the `diacritic_sensitive` would be set to `None`.
https://github.com/mykolaskrynnyk/beanie/blob/0773aed151acb30dfcc4448cdbc26326c7366a53/tests/odm/operators/find/test_evaluation.py#L52C1-L59C6

adeelsohailahmed (approved):
Thank you

mykolaskrynnyk:
@staticxterm , it would've been better if all three parameters had been made optional and set to `None` in the original implementation of `Text` class. Right now, it is a bit of a legacy and consistency issue. `language` is already optional and `None` by default which is fine. `diacritic_sensitive` is made optional in this PR but both `diacritic_sensitive` and  `case_sensitive` are `False` by default. The use case for `diacritic_sensitive=None` is to enable queries agains version 2 text indices (like Azure CosmosDB for MongoDB) but there is no clear-cut use case for explicitly setting `case_sensitive` to `None` except for delegating case sensitivity to server/index defaults, which again should've probably been the default behaviour anyway. To sum up, there are two options:

**Option 1**

I guess this is what you suggest:

```python
# @@ class Text(BaseFindEvaluationOperator):
        search: str,
        language: Optional[str] = None,
        case_sensitive: Optional[bool] = False,
        diacritic_sensitive: Optional[bool] = False,
    ):
```

Which is a non-breaking change but it doesn't seem like there is any use-case for that at this point.

**Option 2**

I think this option would be a lot nicer (and work out of the box for use cases like version 2 text indices) but it does introduce a breaking change:

```python
# @@ class Text(BaseFindEvaluationOperator):
        search: str,
        language: Optional[str] = None,
        case_sensitive: Optional[bool] = None,
        diacritic_sensitive: Optional[bool] = None,
    ):
```

Any additional thoughts on the topic?

staticxterm:
Hi @mykolaskrynnyk,
I agree that it would've been better if all 3 parameters were made optional and set to None from the start.
I guess I was proposing the first option then, keeping all the other parameters as they were with their defaults set to False, just making them all optional (and if they are None, exclude them from the query).
Later on, in a next major version of Beanie, we could I guess make this interface a bit better and make all of these parameters optional/not included, leaving only "search" as mandatory.

Riverfount (commented):
I'll review it after all change requests are solved.

mykolaskrynnyk:
@Riverfount , I don't think there are any outstanding change requests. Can you clarify what you mean?

mykolaskrynnyk:
@Riverfount , if you refer to the change request from @adeelsohailahmed, it has already been solved in subsequent commits and approved accordingly.

PR #1090, opened 2024-12-12
fix: `Owner` model missing in `init_beanie` in Inheritance documentation's Inserts example

ksayer:
Example in documenation is broken, Owner is lost in init


adeelsohailahmed (approved):
Thank you for your contribution!

staticxterm (approved):

PR #1097, opened 2024-12-27
Refactor/index management

mykolaskrynnyk:
This request improves the readability and maintainability of classes and functions related to index management:

- adds a `__hash__` method to `IndexModelField` to allow set operations on collections of indexes.
- removes limited-use `@staticmethod`s from `IndexModelField` in favour of explicit set operations to reduce verbosity.
- enhances the readability of `init_indexes` function by reducing clutter and using self-explanatory variable names.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -186,3 +186,6 @@ tilt_modules
 # Poetry stuff
 poetry.lock
 .pdm-python
+
+# macOS
+**/.DS_Store
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -607,21 +607,14 @@ class IndexModelField:
             )
         )
 
+    def __hash__(self) -> int:
+        return hash(self.fields + self.options)
+
     def __eq__(self, other):
         return self.fields == other.fields and self.options == other.options
 
     def __repr__(self):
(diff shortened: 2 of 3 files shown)


mykolaskrynnyk:
@staticxterm, not sure if this was reopened because you are planning on working on this request or for some other reason. Regardless, here are two main things to note about this pull request:

- The main idea behind the changes was to introduce set operations to simplify index management.
- The glaring omission in the current implementation is the fact that there are two different situations where index comparisons are needed: one where the indices must only match by field names to be considered equal and another one where both field names and options must match.

The current implementation of `__hash__` assumes the latter, which, obviously, does not work for the former:

```python
def __hash__(self) -> int:
        return hash(self.fields + self.options)
```

Basically, there is no one-size-fits-all solution and one would need to implement methods for both types of comparisons and do so neatly. I don't think I'll have the bandwidth to get back to this anytime soon, but if anyone wants to finalise the request, this should be helpful to get started.

staticxterm:
Hi @mykolaskrynnyk,
thank you for asking and for the detailed explanation of this PR.
I am planning to work on #1151 so was looking into this PR and if it could be incorporated to simplify index handling.

PR #1101, opened 2025-01-03
Cache pydantic and mongodb in tests workflow

dantetemplar:
Also format github-actions-tests.yml and remove undefined `mongodb-data-path` param


dantetemplar:
Not yet ready!

PR #1107, opened 2025-01-13
fix BulkWriter.__init__ bug when it has object_class param input

hagd0520:
fix bug occur to two scenarios 
- Initialize `BulkWriter` directly by BulkWriter.__init__ with object_class parm input.
- Initialize `BulkWriter` through `Document.bulk_writer` and Document has collection.

### DIFF
--- a/beanie/odm/bulk.py
+++ b/beanie/odm/bulk.py
@@ -4,8 +4,14 @@ from types import TracebackType
 from typing import TYPE_CHECKING, Any, List, Mapping, Optional, Type, Union
 
 from motor.motor_asyncio import AsyncIOMotorClientSession
-from pymongo import (DeleteMany, DeleteOne, InsertOne, ReplaceOne, UpdateMany,
-                     UpdateOne)
+from pymongo import (
+    DeleteMany,
+    DeleteOne,
+    InsertOne,
+    ReplaceOne,
+    UpdateMany,
+    UpdateOne,
+)
 from pymongo.results import BulkWriteResult
 
 if TYPE_CHECKING:
@@ -78,7 +84,9 @@ class BulkWriter:
         self.object_class = object_class
         self.bypass_document_validation = bypass_document_validation
         self.comment = comment
-        self._collection_name: str = object_class.get_collection_name() if object_class else None
+        self._collection_name: Optional[str] = (
+            object_class.get_collection_name() if object_class else None
+        )
 
     async def __aenter__(self) -> "BulkWriter":
         return self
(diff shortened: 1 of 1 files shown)


hagd0520:
> > fix bug occur to two scenarios
> > 
> > * Initialize `BulkWriter` directly by BulkWriter.**init** with object_class parm input.
> > * Initialize `BulkWriter` through `Document.bulk_writer` and Document has collection.
> 
> the collection name is only used when add_operation is called and doesnt alter anything as long object_class is not none when you commit

hagd0520:
> > fix bug occur to two scenarios
> > 
> > * Initialize `BulkWriter` directly by BulkWriter.**init** with object_class parm input.
> > * Initialize `BulkWriter` through `Document.bulk_writer` and Document has collection.
> 
> the collection name is only used when add_operation is called and doesnt alter anything as long object_class is not none when you commit

bug appears from the add_operation func.

```python
    def add_operation(
        self,
        object_class: Type[Union[Document, UnionDoc]],
        operation: _WriteOp,
    ):
        if self.object_class is None:
            self.object_class = object_class
            self._collection_name = object_class.get_collection_name()
        else:
            if object_class.get_collection_name() != self._collection_name:
                raise ValueError(
                    "All the operations should be for a same collection name"
                )
        self.operations.append(operation)
```

under the scenarios i mentioned, initialized BulkWriter has no `_collections_name` and it raise the ValueError from the line 154 in `bulk.py` cause its object_class is not None and `_collections_name` is None.

so that `_collections_name` should be properly intialized.

staticxterm (approved):

Riverfount (approved):
Good job.

unexceptable:
Can we get a release of Beanie with this fix? Bulk writer (as per the provided docs examples) is broken without this fix, and it feels silly that the fix is still not out given it was merged in Feb :(
 
To avoid pinning to `1.28.0` I'm setting `_collection_name` in my own code:
```
    @classmethod
    async def bulk_update(
        cls,
        update_actions: list[UpdateAction],
        ordered: bool = True,
    ):
        async with cls.bulk_writer(ordered=ordered) as bulk:
            # NOTE(unexcept): Until a new version beyond `1.29.0` is released, we gotta set _collection_name
            #                 see: https://github.com/BeanieODM/beanie/pull/1107
            bulk._collection_name = cls.get_collection_name()
            for update_action in update_actions:
                update = update_action.advanced_update
                if not update:
                    update = { "$set": update_action.update }
                await cls.find_one(update_action.query).update(update, bulk_writer=bulk)
```

JBloss1517:
Thanks for getting this fixed! Just ran into this issue myself.

PR #1117, opened 2025-01-31
Fix: first_or_none no longer mutates FindMany object (#1116)

### LINKED ISSUE #1116: [BUG] After first_or_none of class FindMany, limit == 1 remains
**Describe the bug and reproduce and expected behavior**
After first_or_none of class FindMany, limit == 1 remains.
So 
`python
requests = RequestModel.find() # I have 3 requests 
first_request = await requests.first_or_none() # Return first request
requests_list = requests.to_list() # In my opinion It must to return 3 requests, but returns 1 after first_or_none because first_or_none  assigns limit_number == 1
`

**Solution**
`python
async def first_or_none(self) -> Optional[FindQueryResultType]:
    """
    Returns the first found element or None if no elements were found
    """
    existing_limit_number = self.limit_number
    res = await self.limit(1).to_list()
    self.limit_number = existing_limit_number
    if not res:
        return None
    return res[0]
`

eac0de:
Fixes #1116


beanie/odm/queries/find.py
```diff
@@ -692,10 +692,8 @@ async def first_or_none(self) -> Optional[FindQueryResultType]:
         """
         Returns the first found element or None if no elements were found
         """
-        res = await self.limit(1).to_list()
-        if not res:
-            return None
-        return res[0]
+        result = await self.to_list(1)
```
staticxterm, line 695: While the performance impact may be negligible, I would prefer if the query would be as efficient as possible, so that Mongo only has to return 1 result.
Please write this the old way (use your initial proposal), using `self.limit(1).to_list()` and resetting `self.limit_number` to the previous value after result is retrieved.

staticxterm (changes requested):
Thank you for creating the PR. Just one small comment, please if you can use your original proposal to retain the "limit".

beanie/odm/queries/find.py
eac0de: Sorry, I didn't realise that motor returns the whole list when to_list(1) is called. I've looked into it now and saw that. Thanks, I made an amend to follow the ‘one commit’ rule.

staticxterm (approved):
Looks good. Thank you!

Riverfount (approved):
Good job.

staticxterm:
Hi @eac0de,
this PR has not been merged.
Could you please re-submit this PR?
Thanks.

PR #1123, opened 2025-02-12
Do not ignore deprecation warnings in tests

Viicos:
This will be really useful especially as beanie makes heavy use of Pydantic, **including internal stuff**!


roman-right:
Thank you for this, @Viicos. The approach is solid and the `lazy_model` dependency is merged.

The PR needs a rebase on current `main` though -- a lot changed since it was created (Pydantic v1 removal, py310+ modernization, test restructuring). The conflicts will be significant, especially in `tests/odm/models.py`, `conftest.py`, and `pyproject.toml`.

Could you rebase and push? After that I can run CI and merge.

staticxterm:
@roman-right, I have continued to work on this PR.
It didn't need to be rebased as there were no merge conflicts, but I've rebased on latest main - please review.

PR #1126, opened 2025-02-17
Fixed AttributeError in merge_model when updating Document

HK-Mattew:
Fixed AttributeError in merge_model when updating Document. Close: 1125


tests/odm/models.py
```diff
@@ -627,6 +627,17 @@ class Settings:
         class_id = "123"
 
 
+class DocumentTestModelWithModelConfigExtraAllow(Document):
+
+    class Settings:
+        use_cache = True
+        cache_expiration_time = datetime.timedelta(seconds=10)
+        cache_capacity = 5
+        use_state_management = True
```
staticxterm, line 636: In my opinion, the settings can be removed as they're not necessary. Lets keep this model as minimal as possible.

tests/odm/models.py
```diff
@@ -627,6 +627,17 @@ class Settings:
...
+
+    model_config = ConfigDict(extra='allow')
```
staticxterm, line 638: For the time being, while we do still support Pydantic v1, this should be wrapped inside an if-else block. For an example, please see lines 573-580.

tests/odm/documents/test_update.py
```diff
@@ -127,6 +128,14 @@ async def test_update_one(document):
     assert new_document.test_list[0].test_str == "foo_foo"
 
 
+async def test_update_one_set_extra_field():
```
staticxterm, line 131: This will make tests fail on Pydantic v1 because the `DocumentTestModelWithModelConfigExtraAllow` model uses Pydantic v2 style config only. Please see my other comment for the `DocumentTestModelWithModelConfigExtraAllow` class.

staticxterm (changes requested):
Hi @HK-Mattew, thank you for the PR. Great to see PRs that add better support for Pydantic v2!
Adding some preliminary comments, I'll test this out more soon.

tests/odm/models.py
HK-Mattew: You are right. I just copied an already existing template and kept it that way.

tests/odm/models.py
HK-Mattew: It's fixed. I had forgotten about pydantic v1 :/

beanie/odm/utils/parsing.py
```diff
@@ -24,7 +24,7 @@ def merge_models(left: BaseModel, right: BaseModel) -> None:
     from beanie.odm.fields import Link
 
     for k, right_value in right.__iter__():
-        left_value = getattr(left, k)
+        left_value = getattr(left, k, None)
```
staticxterm, line 27: Such a simple fix, I like it.
Easy to overlook.

staticxterm (approved):
Looks good, short and sweet, thank you!

beanie/odm/utils/parsing.py
HK-Mattew: A simple solution that solves a big problem :)

Riverfount (approved):
Great job.

HK-Mattew:
Only one person left to review. Anyone?

binaryarun (approved):

PR #1133, opened 2025-02-24
fix: do not override Document._class_id if it was explicitly set on the document class

### LINKED ISSUE #1132: Document._class_id is always overwritten
**Describe the bug**

`Document._class_id` is overwritten. Document initialization should check if it's set before assigning a value. This messes up the customization of document discriminators and leads to failing queries (update, find, link fetching...).

**To Reproduce**

Create a couple of documents (with inheritance pattern), set `Settings.class_id` for example to `"type"`, add `type: str = "my-doc-type-1"`, `type: str = "my-doc-type-2"`, ... properties to documents and also set `_class_id` on each document to the same value as the `type` property.

**Expected behavior**

Update, find, link fetching queries work.

**Additional context**

I'll submit a PR with a fix that avoids overwriting `_class_id` if it's already set. It solves the issues.

volfpeter:
Fixes #1132

### DIFF
--- a/beanie/odm/utils/init.py
+++ b/beanie/odm/utils/init.py
@@ -376,7 +376,8 @@ class Initializer:
         cls._children = dict()
         cls._parent = None
         cls._inheritance_inited = False
-        cls._class_id = None
+        if not hasattr(cls, "_class_id"):
+            cls._class_id = None
         cls._link_fields = None
 
     @staticmethod
@@ -598,11 +599,13 @@ class Initializer:
                     class_name=cls.__name__,
                     collection_name=cls.get_collection_name(),
                 )
-                cls._class_id = cls.__name__
+                if cls._class_id is None:
+                    cls._class_id = cls.__name__
                 cls._inheritance_inited = True
             elif output is not None:
                 output.class_name = f"{output.class_name}.{cls.__name__}"
-                cls._class_id = output.class_name
+                if cls._class_id is None:
+                    cls._class_id = output.class_name
                 cls.set_collection_name(output.collection_name)
                 parent.add_child(cls._class_id, cls)
                 cls._parent = parent


staticxterm:
Hi @volfpeter. Thank you for the issue report/PR.
Please also write tests for the scenarios that you described in the issue report, to have failing tests that test and reproduce this behavior prior to the changes made here in this PR.

volfpeter:
Hi, Thanks for the quick response. I just fixed the existing tests. I'll add a new test for this specific case.

volfpeter:
@staticxterm Do you have an idea why the "3.13, 8.0, 1.10.18" test failed? I ran the test suite locally with the same Python and package versions (Python 3.13, pytest 8.3.4, and both pydantic 1.10.18 and 1.10.21), and all tests passed. Flaky tests?

staticxterm:
Hi @volfpeter, the test failed because yes there is some flakiness. I'll look into it.
I'll rerun the failed test.

volfpeter:
Thanks! First I thought this was the case, but then it seemed that locally it's always passing, and the CI job is always failing, and I didn't want to trigger to useless reruns.

volfpeter:
Hi guys! Is there any chance of a review?

staticxterm:
Hi @volfpeter. I have looked more closely into this PR. 
"Document._class_id" is a Document class internal field. The only reference in Beanie documentation for its usage is here: https://beanie-odm.dev/tutorial/multi-model-pattern/
> _class_id is the default beanie internal field used to filter children Documents
But also from the code it can be inferred that it is used for the Inheritance feature.
https://github.com/BeanieODM/beanie/blob/main/beanie/odm/interfaces/inheritance.py

Beanie sets and uses this field internally. Reviewing the current Beanie documentation, there is no indication that this field should be modifiable. I am sorry, but I believe that it should not be changed.

What are your use cases for wanting to change this? I am not seeing the practical benefit of this change from the provided tests.

volfpeter:
Hi,

We use the multi-model pattern for a collection, and our discriminator field is called `type`, whose value comes from a literal type (just a static set of strings). We need this field to have static values, dynamically created class IDs like `BaseDocument.AnotherBaseDocument.ActualBusinessObject` won't do (changing class names or inheritance would break the app, because class IDs would change).

Our attempted solution to this problem was to add `class_id = "type"` to our base document's `Settings` (in addition to `is_root = True`), and then we manually set `_class_id` in each leaf document to the necessary value. This did not work at all, because the already set `_class_id` was always overwritten by beanie. This bug is fixed by the PR.

The PR changes only one thing: if `_class_id` is set on a document class, use that value instead of overwriting it with an automatically generated one. If `_class_id` is not set, then for that class the existing behavior is fully preserved (including "namespacing" with parent class `_class_id`s). I can't see how doing anything else could be correct.

I assume the main reason for `_class_id` to be prefixed with an underscore is to never accidentally include it in the [Pydantic schema](https://docs.pydantic.dev/latest/concepts/models/#class-variables) and to not have potential name conflicts with users' code.

If there is an already working solution to this use-case, I'd love to know, but it doesn't seem to be documented.

(Footnote: actually, setting the custom class ID in `Settings` would be a better solution, but my goal was to do the minimum necessary change.)

I hope I managed to change your mind about accepting this bugfix.

volfpeter:
> Beanie does allow to change/specify the custom "class ID" field but then why isn't its value used should be checked/fixed

I assume you're referencing `Settings.class_id`? The value of that setting must be a model field's _name_, something that's in the DB. It can not be used to determine the value of `_class_id`, because it's a model field, not a beanie configuration attribute, and it may not have a default value.

> instead of overriding this IMO internal/private field

My point is still that the user declares the document class, they can set `_class_id` if they want a fixed value that's not derived from the class name (and inheritance chain). And if the user does this, then beanie shouldn't override it during initialization. Other solutions just add extra complexity. For example another setting that stores the class ID itself or the name of the class attribute that stores it. Actually, `class_id` would be a good setting name for this, but it's already used and the current interpretation of it is `class_id_field` (This is a bit confusing, but maybe it's just me and my relative lack of experience with beanie).

(FWIW: this fix is already working nicely in production and is fully backward compatible. And it would also not interfere with potential future changes to how the class ID value should be set.)

Anyway, thanks for taking a look at this!

volfpeter:
Hi, have you managed to make progress on this?

volfpeter:
@staticxterm @roman-right Sorry for pushing it, but could you please review this PR? The bug report and the fix have been waiting since February. I'm happy to make changes if required, just let me know. Thank you!

staticxterm:
Hi @volfpeter,
sorry for the late response, I haven't had the time yet to look into this in more detail, mainly, whether this is a good thing to allow to override.
Regarding what you mentioned in one previous comment:
> (Footnote: actually, setting the custom class ID in Settings would be a better solution, but my goal was to do the minimum necessary change.)

I would agree with you here and I am still of the opinion that `Document._class_id ` is a protected variable, set internally by Beanie, and should not be allowed to be set/redefined on sub-models of the base Document.
So I propose that we ~~reuse the `class_id`~~ introduce a new configuration option in the `Settings` subclass and add functionality to override it for the child Documents, also by setting it through the `Settings` subclass.

I am of the opinion that the current current way that the "class_id" setting is defined and used is bad design and confusing at the very least. There is also no documentation that this field could be used to change the default Document "key" that Beanie internally sets and uses for the inheritance/multi-model pattern. 

We should also add documentation for this once we make the changes.

tldr; `cls._class_id` is something that is set internally by Beanie and should not be changed. We should look for other ways to configure it.

volfpeter:
> I haven't had the time yet to look into this in more detail, mainly, whether this is a good thing to allow to override. 

It would be nice if you could actually have a look at the code and properly think through the fix and the implications of alternative implementations, before I start working on a totally different solution!

As I described earlier, I still think `_class_id` starts with an underscore to always omit it from the Pydantic schema (otherwise users would always need to add the `ClassVar` annotation which they would usually forget), and I also still think blindly overwriting `_class_id` during model initialization is a bug (fixed by this PR) in itself. Whether the user setting `_class_id` to configure the discriminator *value* (not field name) is the right "public" API is a different question (and a missing feature).

I thought about this a bit more, and I actually have a quite important counter-argument for the potential new setting:

*Settings are inherited*, but this new setting obviously *must not be inherited*. This will be confusing to users. Having another class variable instead of a setting would fix this issue, but then again, the variable would need to start with an `_`, and in that case, why don't we just use `_class_id`?

If you still think a setting would be the right way to go, what should be its name (reminder: this would store the discriminator value, not the field name)? The correct name for the current `Settings.class_id` would be `class_id_field` (because it is the *name of the field* that stores the subclass discriminator, and as such, it must not be overwritten in subclasses), but changing it and breaking the API is obviously not recommended. So `class_id_val` or `class_identifier`? Quite ugly or confusing.

It would be really nice to clarify all these question before I put more work into this. Or, alternatively, if the PR is accepted as a bugfix, then I could work on the "feature" part in a separate PR (including docs and everything). But I'd really like to finally get to the end of this bug.

@staticxterm @roman-right Looking forward to your feedback. Thanks for your time!

staticxterm:
Hi @volfpeter , regarding your first part of the reply.
I did, and I still stand by what I wrote in every reply. I am not for allowing the usage/override of internal/private/protected fields.
Even if it's just for some corner case, it has the potential to spread,, it's bad design and would need to be maintained for the long run.
It's better to do this right the first time.
I leave all other potential questions for you or others to agree upon - e.g. for naming,  I have no preferences as long as it's clear and readable.

volfpeter:
> I leave all other potential questions for you or others to agree upon - e.g. for naming, I have no preferences as long as it's clear and readable.

Okay then. I'll start working on the `Settings.class_id_val` option, still in this PR if you don't mind, since most of the code change is tests, and those would remain essentially the same.

PR #1134, opened 2025-02-28
Support partial list update (append) in state manager

bushig:
### What's New

- Added smarter handling of array operations with state management. Now when you append to arrays, Beanie uses MongoDB's `$push` operator rather than replacing the entire array. 
- Also some changes to docs to make cleaner why one would use state manager.
- fixture to check what operations was actually sent to Mongo by motor, instead of using get_changes.

### More details
- Array `append()` operations now generate MongoDB `$push` operations  with  `state_management=True`
- Only happens when `state_management_replace_objects=False` (which is the default)
- Other array operations (changing items, removing them) still work as before, cause they unsafe for sure.
- The `get_changes()` method still shows the full array for compatibility

Also thinking about adding option for unsafe direct array operations like changing elements by index, but still evaluating if its worth it.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,9 @@
+from typing import List, Optional, Set
+
 import motor.motor_asyncio
+import pymongo.monitoring
 import pytest
+from pymongo.monitoring import CommandListener
 
 from beanie.odm.utils.pydantic import IS_PYDANTIC_V2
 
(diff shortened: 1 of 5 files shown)


bushig:
Fixed tests, but there is still two failing because of docker hub image pull limits and one seems to be [flaky test](https://github.com/BeanieODM/beanie/actions/runs/13630429494/job/38096850773?pr=1134)
Can anyone rerun failing tests?

staticxterm:
@BeanieODM/beanie-devs is anyone willing to take a look at this PR? Although e.g. the data loss issue may still occur due to concurrent update from other processes, I believe this feature and the doc updates are worth merging.

roman-right:
Using \ instead of replacing entire arrays on append is a nice optimization for state management. The mergeability is unknown though, so this likely needs a rebase. Also, the pre-commit check failed - please fix that too. If you are still interested, please update and I will review.

tests/conftest.py
```diff
@@ -19,8 +23,49 @@ def settings():
...
+        A configured CommandLogger instance
+    """
+    logger = CommandLogger({"findAndModify", "update", "insert"})
+    pymongo.monitoring.register(logger)
+    yield logger
+
+
 @pytest.fixture()
-def cli(settings):
+def cli(settings, command_logger):
```
roman-right, line 68: Making `cli` depend on `command_logger` means every test that uses `cli` (almost all of them) will call `pymongo.monitoring.register()`. These listeners are global and cannot be unregistered, so by the end of a test session there will be hundreds of active listeners receiving every MongoDB command. This will slow down the entire test suite. Consider keeping `command_logger` as an independent fixture, only used by the tests that actually need it.

roman-right (changes requested):
The $push optimization for list appends is a good idea. One concern about the test infrastructure change that would affect the full test suite.

PR #1139, opened 2025-03-13
fix find type hint for filter args

CAPITAINMARVEL:


Riverfount (approved):
LGTM

staticxterm (approved):
Didn't check / test, but @CAPITAINMARVEL does it really fix anything or just silences the warnings? Since "Any" means we do not know the actual type or that it can be "anything".

07pepa (approved):

PR #1141, opened 2025-03-16
Fix failing tests in CI

staticxterm:
Try to fix test flakiness in CI.
Minor change in non-test code where the Motor client is now also being closed.


Riverfount (approved):
Great job.

07pepa (approved):

PR #1144, opened 2025-03-20
Update toml dependency to use tomli instead of toml

Taaku18:
The toml library beanie uses is years out of date, and [doesn't support the newer toml syntax (from 5 years ago...) of mixing types in arrays](https://github.com/uiri/toml/issues/270). 

When running migrations, beanie uses the toml library to parse config from pyproject.toml. If the pyproject file contains mixed types, such as this valid tox config:

```
[tool.tox.env.precommit]
description = "run pre-commit"
skip_install = true
deps = ["pre-commit"]
commands = [["pre-commit", { replace = "posargs", default = ["run", "--all-files"], extend = true} ]]
```

`beanie migrate -p [...]` will error out with:

    File ".venv/lib/python3.12/site-packages/toml/decoder.py", line 1002, in load_array
      a[b] = a[b] + ',' + a[b + 1]
                          ~^^^^^^^
    IndexError: list index out of range

This PR changes the dependency from toml to tomli (and tomli-w for GH script). It's the most used toml library that provides backports of the builtin tomllib to py3.8.

### DIFF
--- a/.github/scripts/handlers/version.py
+++ b/.github/scripts/handlers/version.py
@@ -2,7 +2,8 @@ import subprocess
 from pathlib import Path
 
 import requests  # type: ignore
-import toml
+import tomli
+import tomli_w
 from gh import GitHubHandler
 
 
@@ -57,7 +58,8 @@ class VersionHandler:
 
     @staticmethod
     def parse_version_from_pyproject(pyproject: Path) -> SemVer:
-        toml_data = toml.loads(pyproject.read_text())
+        with pyproject.open("rb") as f:
+            toml_data = tomli.load(f)
         return SemVer(toml_data["project"]["version"])
 
     def get_version_from_pypi(self) -> SemVer:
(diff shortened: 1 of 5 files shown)


staticxterm (changes requested):
Hi, thank you for the PR. Two suggestions:

- specify and pin down the dependency versions
- don't use tomli for Python 3.11 and greater

Taaku18:
> Hi, thank you for the PR. Two suggestions:
> 
>     * specify and pin down the dependency versions
> 
>     * don't use tomli for Python 3.11 and greater

I've updated the dependency list to include the version locks as requested.
Also switched to importlib for >=3.11; used `sys.version_info` check instead of your suggested `try...except ImportError...` to satisfy mypy.

staticxterm (approved):
Looks good. Thank you.

Riverfount (approved):
Great job.

PR #1149, opened 2025-04-05
use UV as dependency management tool

07pepa:


.github/workflows/github-actions-tests.yml
```diff
@@ -15,27 +15,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12", "3.13" ]
+        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "3.13t" ]
```
07pepa, line 18: "3.13t" is no gil version

.github/workflows/github-actions-tests.yml
```diff
@@ -15,27 +15,29 @@ jobs:
...
-      - name: install dependencies
-        run: pip install .[test,ci]
-      - name: install pydantic
-        run: pip install pydantic==${{ matrix.pydantic-version }}
+      - name: install dependencies for pydantic1
+        if: matrix.pydantic-version == '1'
+        run: uv sync --group test  --group ci --frozen && uv add "pydantic[email]==1.10.18" pydantic-extra-types
+      - name: install pydantic2 deps
+        if: matrix.pydantic-version == '2'
+        run: uv sync --group test  --group ci --frozen && uv add "pydantic[email]==2.11.3" pydantic-settings pydantic-extra-types
```
staticxterm, line 39: `uv sync --group test  --group ci --frozen` actually installs all other dependencies so the description/name "install dependencies for pydantic1" / "install pydantic2 deps" may be a bit misleading here...
Yes, dependencies for Pydantic v1 are being installed here along with all other test dependencies.
My suggestion is to perhaps split it into 3 steps. First install all "test" dependencies, then install either Pydantic v1 or Pydantic v2 dependencies.

```suggestion
      - name: Install test dependencies
        run: uv sync --group test --group ci --frozen
      - name: Install Pydantic v1 and its dependencies
        if: matrix.pydantic-version == '1'
        run: uv add "pydantic[email]==1.10.18" pydantic-extra-types
      - name: Install Pydantic v2 and its dependencies
        if: matrix.pydantic-version == '2'
        run: uv add "pydantic[email]==2.11.3" pydantic-settings pydantic-extra-types
```

.github/workflows/github-actions-tests.yml
```diff
@@ -15,27 +15,29 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12", "3.13" ]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.13t" ]
```
staticxterm, line 18: Keep it consistent, since a space was removed from the beginning.

```suggestion
        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.13t"]
```

.github/workflows/github-actions-tests.yml
```diff
@@ -15,27 +15,29 @@ jobs:
...
         mongodb-version: [ "4.4", "5.0", "6.0", "7.0", "8.0" ]
```
staticxterm, line 19: No spaces around to keep it consistent.
```suggestion
        mongodb-version: ["4.4", "5.0", "6.0", "7.0", "8.0"]
```

.github/workflows/github-actions-tests.yml
```diff
@@ -15,27 +15,29 @@ jobs:
...
-        pydantic-version: [ "1.10.18", "2.9.2" , "2.10.4"]
+        pydantic-version: [ "1", "2"]
```
staticxterm, line 20:
```suggestion
        pydantic-version: ["1", "2"]
```

.github/workflows/github-actions-tests.yml
```diff
@@ -15,27 +15,29 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
```
staticxterm, line 23: If we look in the "run tests" stage output for every pipeline "run tests" job, we can see that the Python version used is always "Python 3.12.3", so the setting of the Python version is not working (since it was removed).
I've added back the "setup-python" stage:
```suggestion
      - uses: actions/checkout@v4
      - name: Set up Python ${{ inputs.python-version }}
        uses: actions/setup-python@v5
        with:
          python-version: ${{ inputs.python-version }}
```

.github/workflows/github-actions-tests.yml
```diff
@@ -15,27 +15,29 @@ jobs:
...
-      - uses: actions/setup-python@v5
+      - name: Install uv and set the python version
```
staticxterm, line 24: As explained in previous comments, the Python version was not being set here in this step...

```suggestion
      - name: Set up uv
```

staticxterm (changes requested):
Honestly, it looks very very good, I like it.
Needs some minor tweaks to fix the Python version used in the CI tests and some other comments.

pyproject.toml
```diff
@@ -42,9 +42,6 @@ test = [
     "asgi-lifespan>=1.0.1",
     "httpx>=0.23.0",
     "fastapi>=0.100",
-    "pydantic-settings>=2",
-    "pydantic-extra-types>=2",
-    "pydantic[email]",
```
staticxterm, line 47: I have another issue when trying to run `pytest` locally due to this.
I think this should be reverted and stay here. Then in the CI tests we can install the wanted versions of these dependencies (upgrade or downgrade).
This is the error I get:
```bash
$ uv run pytest -vvv
ImportError while loading conftest '<REDACTED>\Projects\beanie\tests\conftest.py'.
tests\conftest.py:7: in <module>
    from pydantic_settings import BaseSettings
E   ModuleNotFoundError: No module named 'pydantic_settings'
```

staticxterm (changes requested):

Riverfount (dismissed):
Great job.

.pre-commit-config.yaml
```diff
@@ -13,3 +13,8 @@ repos:
           - types-click
           - types-toml
         exclude: ^tests/
+
```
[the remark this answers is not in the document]
07pepa: causes issues lets do it in separate pr

.pre-commit-config.yaml
```diff
@@ -13,3 +13,8 @@ repos:
...
+  - repo: https://github.com/astral-sh/uv-pre-commit
+    rev: 0.6.12
```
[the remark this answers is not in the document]
07pepa: they are fast as hell

pyproject.toml
07pepa: yeah  this is hard to do... even as optional dep.... pydantic_settings only supports pydantic>=2 as far as i know so we need to drop pydantic 1 to do this

07pepa: as long as pydantic_settings are in any dependencies they become unsatisfiable

staticxterm (changes requested):
Like I mentioned previously, I like this.
But, I can't yet approve it as there are still some comments left unanswered and the development experience needs to be improved (namely that doing uv sync.. installs all needed development and test dependencies - that everything just works out of the box).

pyproject.toml
07pepa: @staticxterm  i would go major bump along with removing python 3.8

07pepa: and keeping pydantic 1.x would result in confusion ci cd just hard adds it at particular version as workaround 

doing it localy would add it to uv.lock and make live misserable

PR #1150, opened 2025-04-08
Fix PydanticDeprecatedSince211: Accessing the 'model_fields' attribute on the instance is deprecated

gsakkis:


cdelacruzpinto:
This would fix https://github.com/BeanieODM/beanie/issues/676, which even being closed, is still present on 1.29.

staticxterm (changes requested):
I tested this locally and I still get the warning in question. This is due to Beanie having locked lazy-model to version 0.2.0 in pyproject.toml:
https://github.com/BeanieODM/beanie/blob/ea651add303c2e512eab54590ca6ff3542d55cd7/pyproject.toml#L30

Please also uplift the lazy-model dependency to version 0.3.0, which should finally fix this warning once and for all.

Currently the warnings are not visible in our CI since they are being suppressed by this line:
https://github.com/BeanieODM/beanie/blob/ea651add303c2e512eab54590ca6ff3542d55cd7/pyproject.toml#L102

JBorrow:
Fix looks good and will reduce noise in our application and CI.

martincpt:
Will this be merged? And if so, will the 1.x branch receive it? I get around 115,000 warnings when running my tests, which is way too many.

staticxterm:
Hi @martincpt, thank you for raising this issue. I'll assign myself to this outstanding issue #676 since we need to fix all warnings/errors from inside the Beanie tests scope as well (currently the warnings are being suppressed so they are not visible in our CI).
Unfortunately I don't know how to backport the fixes for the 1.x branch so they will only be contained in the 2.x version of Beanie...

martincpt:
Hi @staticxterm, theoretically, can't we just cherry pick this commit on the 1.x branch?

PR #1151, opened 2025-04-11
Index merging improvements

mikelei8291:
I was trying to create two indexes on the same field for my model:

```python
class Order(OrderInit, Document):
    created_at: Annotated[
        AwareDatetime,
        Field(default_factory=lambda: datetime.now(UTC))
    ]
    status: Annotated[OrderStatus, Indexed()]

    class Settings:
        indexes = [
            IndexModel(
                "created_at", name="processing_expire_30", expireAfterSeconds=30, partialFilterExpression={"status": "processing"}
            ),
            IndexModel(
                "created_at", name="unpaid_expire_900", expireAfterSeconds=900, partialFilterExpression={"status": "unpaid"}
            )
        ]
```

However, the current implementation will overwrite the first index definition with the second one because they have the same dict key if the key is `index.fields`. I changed the key to `index.name` so by providing a different name than the default one, the indexes will no longer overwrite each others.

Also I was trying to make it work with `Annotated` too:

```python
class Order(OrderInit, Document):
    created_at: Annotated[
        AwareDatetime,
        Indexed(name="processing_expire_30", expireAfterSeconds=30, partialFilterExpression={"status": "processing"}),
        Indexed(name="unpaid_expire_900", expireAfterSeconds=900, partialFilterExpression={"status": "unpaid"}),
        Field(default_factory=lambda: datetime.now(UTC))
    ]
    status: Annotated[OrderStatus, Indexed()]
```

but it would change the return type of [`get_index_attributes()`](https://github.com/BeanieODM/beanie/blob/f19416dbfc1c539770dab095f05b2bc18cb73741/beanie/odm/utils/typing.py#L28) so I didn't go further.

### DIFF
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -1,4 +1,6 @@
 import asyncio
+import json
+import logging
 from collections import OrderedDict
 from dataclasses import dataclass
 from enum import Enum
@@ -39,6 +41,15 @@ from beanie.odm.utils.pydantic import get_model_fields
 
 if TYPE_CHECKING:
     from beanie.odm.documents import DocType
+
+logger = logging.getLogger(__name__)
+
+_DIRECTION_NAMES = {1: "ASCENDING", -1: "DESCENDING"}
+
+
+def _direction_name(direction: int) -> str:
+    """Return a human-readable name for a PyMongo direction constant."""
+    return _DIRECTION_NAMES.get(direction, str(direction))
 
 
 @dataclass(frozen=True)
@@ -585,7 +596,6 @@ class IndexModelField:
     def __init__(self, index: IndexModel):
         self.index = index
         self.name = index.document["name"]
-
         self.fields = tuple(sorted(self.index.document["key"]))
         self.options = tuple(
             sorted(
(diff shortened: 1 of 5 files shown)


adeelsohailahmed (commented):
Hi @mikelei8291, thanks for spotting and fixing this bug. Your implementation looks sound to me. Could you please add a test for this case too?

staticxterm (dismissed):
Hi @mikelei8291, thank you for the PR.

staticxterm:
I am still working on this. It's tricky. Sometimes we want to merge the indexes, sometimes not.

Riverfount (dismissed):
LGTM

roman-right:
The problem is real - multiple indexes on the same field with different options (like different TTL with partial filters) should not overwrite each other. Using index name as the dict key instead of field name is a reasonable fix. The PR is large though (697 additions). Tests pass. Please rebase against current main if you are still interested, and I will do a thorough review.

roman-right:
@Riverfount Sorry, I accidentally dismissed your review while testing an automation tool. Could you please re-review when you get a chance? Apologies for the inconvenience.

staticxterm:
@roman-right, I am still working on this. Hopefully I'll be able to continue with it once I get some free time.
Currently what I'm struggling with when indexes should be merged, i.e. old ones dropped, then what index should be created if there are aliases on some field... Those kind of things, so it requires extensive testing.

mikelei8291:
@staticxterm Sorry for pinging you. Is there any chance that you will continue working on this fix? Thank you.

staticxterm:
Hi @mikelei8291, I've been focusing on Pydantic V2 aliases support but I've hit a 'dead end' there (too many breaking changes..). I may pick this up and check the state of this tomorrow, thank you for reminding me.
Best regards

mikelei8291:
Oh no :cry:

Issue #756, opened 2023-10-20
Pydantic "exclude" option is not working anymore


arielschvartz:
**Describe the bug**
After upgrading from version 1.22.6 to version 1.23.0, pydantic property exclude stopped working.

**Expected behavior**
I have a user model which have email and password and I don't want to save the password to mongo. I have the following model:
```
class UserCreate(User):
    """All fields needed to create a user"""
    password: Optional[str] = pydantic.Field(
        description="A user's password in plain text before it has been hashed",
        exclude=True,
    )
```
I expected the password not to be saved into the database but after upgrading versions it is saving it.

gsakkis:
I'm not sure which commit caused this but could you try if [this PR](https://github.com/roman-right/beanie/pull/741) happens to fix it?

roman-right:
Hi @arielschvartz,
Thank you for catching that. Did you get a chance to check what @gsakkis asked? I'll address this during the next bug-fixing session next week.

arielschvartz:
Hi there @gsakkis and @roman-right . Sorry for the delay here.

I have not tested specifically this PR, but I tested upgrading to 1.23.1 (as I understood the PR was merged). Unfortunately, it still doesn't look it was fixed. I can do some more specific testing next week if needed.

arielschvartz:
@gsakkis  @roman-right 
Just to confirm, I've tested the new version 1.23.6 and still has the problem.

TheBloke:
Is there any ETA for this? I too want to use exclude to omit fields from Mongo

mhdzumair:
+1

slingshotvfx:
+1, just ran into this today with beanie 1.29.0.

I'd like to be able to `exclude=True` deprecated fields so that it's okay if they exist in the database but don't get written back in the future.

staticxterm:
Right, so the problem is either in the `_iter_model_items()`
https://github.com/BeanieODM/beanie/blob/a2f14d5c39516aa52a95a0c146a9e273be761dbe/beanie/odm/utils/encoder.py#L160
or in the `get_dict()`  methods (or both).
https://github.com/BeanieODM/beanie/blob/a2f14d5c39516aa52a95a0c146a9e273be761dbe/beanie/odm/utils/dump.py#L21
The Pydantic FieldInfo (or any better way to extract the "exclude" flag) should be checked there and excluded fields removed.
Thank you all for reporting this, PRs are welcome.

PR #1154, opened 2025-04-17
fix: Pydantic "exclude" option is not working #756

CatBraaain:
This PR fixes an issue where Beanie document models did not respect the `exclude=True` option from Pydantic.  
Previously, even fields marked as excluded were being saved to the database.  
This behavior is now corrected.
A test case is also added.

Related issue: https://github.com/BeanieODM/beanie/issues/756


07pepa (approved):

beanie/odm/utils/dump.py
```diff
@@ -16,9 +16,12 @@ def get_dict(
         exclude = set()
     if document.id is None:
         exclude.add("_id")
-    if not document.get_settings().use_revision:
-        exclude.add("revision_id")
-    encoder = Encoder(exclude=exclude, to_db=to_db, keep_nulls=keep_nulls)
+    include = set()
+    if document.get_settings().use_revision:
+        include.add("revision_id")
```
staticxterm, line 21: These changes don't seem necessary for the bug fix since "explicitly not excluded fields should by default be included".

beanie/odm/utils/encoder.py
```diff
@@ -83,6 +83,7 @@ class Encoder:
     """
 
     exclude: Container[str] = frozenset()
+    include: Container[str] = frozenset()
```
staticxterm, line 86: Additional class variable does not seem necessary.
Also, there is no test case showcasing its usage (in my opinion, it is not needed as it complicates the design).

beanie/odm/utils/encoder.py
```diff
@@ -154,15 +155,39 @@ def encode(self, obj: Any) -> Any:
...
+    def _should_exclude_field(
+        self, key: str, field_info: Optional[pydantic.fields.FieldInfo]
+    ):
+        exclude, include = (
+            self.exclude,
+            self.include,
+        )
+
+        if key in include:
+            return False
```
staticxterm, line 178: Shouldn't checking only for the existence of "Field(exclude=True, ...)" / "field_info.exclude" be enough?
If it is not excluded, it will by default be included.

beanie/odm/utils/encoder.py
```diff
@@ -154,15 +155,39 @@ def encode(self, obj: Any) -> Any:
...
+        is_pydantic_excluded_field = (
+            field_info is not None
+            and (
+                field_info.exclude
+                if IS_PYDANTIC_V2
+                else getattr(field_info.field_info, "exclude")
+            )
+            is True
+        )
+        return key in exclude or is_pydantic_excluded_field
```
staticxterm, line 189: This check should be enough and the only thing that needs to be evaluated.

staticxterm (changes requested):
Looks good but the additional `Encoder.include` class variable is not necessary.

beanie/odm/utils/dump.py
CatBraaain: The `revision_id` field has `exclude=True` set explicitly:
```python
# https://github.com/CatBraaain/beanie/blob/75684208c0f8dc2fb00447cac9954ee52cbfa2b2/beanie/odm/documents.py#L155
class Document(...)
    ...
    revision_id: Optional[UUID] = Field(default=None, exclude=True)
```
Because of this, even when document.get_settings().use_revision == True, the revision_id field will not be included by default in Encoder class.

beanie/odm/utils/encoder.py
CatBraaain: As mentioned earlier, the revision_id field will not be included by default.
But, when document.get_settings().use_revision == True, revision id is needed to be included.

beanie/odm/utils/encoder.py
CatBraaain: I agree that a test case demonstrating the usage should be added.
I also agree that this addition introduces unnecessary complexity.

I believe the complexity stems from the need to support Pydantic v1 compatibility.
However, looking at PR #1170, it seems there's ongoing work toward removing Pydantic v1 support.
Once that removal is complete, I believe this can be simplified significantly by leveraging Pydantic’s built-in model_dump() method.

beanie/odm/utils/encoder.py
CatBraaain: If the suggestion is to inline the condition without using is_pydantic_excluded_field, I would prefer to keep the intermediate variable because this condition spans multiple lines, using a well-named intermediate variable helps improve readability and understandability of the code.

CatBraaain (commented):
Thanks for the detailed feedback.

I agree that the added complexity isn't ideal, and I also agree that a usage test case should be included if we decide to keep this logic.

However, the `revision_id` field is explicitly marked as `exclude=True`, so it won't be included by default even when `use_revision` is enabled. Therefore, the inclusion must be handled explicitly.

That said, I believe this can be greatly simplified once Pydantic v1 support is dropped (per PR #1170), by using `.model_dump()` from Pydantic v2.
How about we deferring this change until after the Pydantic v1 removal, and revisiting it using the modern API?

beanie/odm/utils/encoder.py
staticxterm: No suggestion here, keep as-is, just wanted to highlight the importance of this check.

beanie/odm/utils/encoder.py
staticxterm: IMO, this is the expected behavior. `revision_id` shall only be included when needed (when `cls.settings.use_revision = True`). Since it's defined on the Document class level, we exclude it here.

beanie/odm/utils/encoder.py
staticxterm: I disagree that this added complexity of `cls./self.included` is needed due to Pydantic v1 support. It worked before (`revision_id` was being excluded as expected), right?
Regarding support for `model.model_dump()` from Pydantic v2, we would need to be very careful when introducing this. Roman introduced this kind of a 'facade' pattern for a reason, since we handle the 'serialization' of the Document model in a special way, abstracting away what Pydantic also does, but extended with types that are needed and specific to Beanie (e.g. ObjectId).

My last thought on this `cls./self.include` is, please, remove it as it is not necessary.

beanie/odm/utils/dump.py
staticxterm: This is as expected, and does not rely on the Pydantic v2 exclusion system, but on the one defined by Beanie (` Encoder(exclude=exclude, ...).encode(...)`).

CatBraaain: The `revision_id` field is defined with `Field(exclude=True)`, and due to the changes in this PR, fields explicitly marked with `exclude=True` in the Pydantic model are now always excluded in `Encoder` class.

Previously, `Encoder(exclude=...)` was used declaratively to control which fields to exclude. However, now `Field(exclude=True)` takes precedence, and the field will be excluded regardless of the `exclude` option passed to the `Encoder`.

Therefore, if we want to include `revision_id` only when `document.get_settings().use_revision == True`, the previous behavior—where it would be included unless explicitly excluded—no longer holds. It is now necessary to explicitly specify `include={"revision_id"}`.

If my explanation is unclear, I'm happy to modify the code as suggested and demonstrate that the existing tests will fail, in order to clarify the behavior.

beanie/odm/utils/encoder.py
CatBraaain: The reason I'm suggesting a shift towards using `.model_dump()` is that it already supports both `include` and `exclude` options natively, allowing us to delegate custom exclude/include logic to the built-in method provided by Pydantic v2.

I understand that in Pydantic v1, such capability was not available — there was no method with support for field-level `include`/`exclude`, which made it necessary for Beanie to implement its own logic at that time.

However, now that Pydantic v2 provides this feature out of the box, and importantly, `.model_dump()` respects fields marked with `exclude=True` as well as computed fields, I believe we can simplify our code significantly by leveraging `.model_dump()` instead of maintaining our own custom filtering logic.

beanie/odm/utils/dump.py
staticxterm: Sorry @CatBraaain but your reasoning here is invalid:
> Therefore, if we want to include revision_id only when document.get_settings().use_revision == True, the previous behavior—where it would be included unless explicitly excluded—no longer holds. It is now necessary to explicitly specify include={"revision_id"}.

Yeah, revision_id has the Field(exclude=True, ...), and in the previous version it was also by default added to the exclude list since Beanie did not have support for the Pydantic "exclude" option (you can say the whole Encoder class is Beanie's own way of re-implementing this).
But your reasoning that it was previously included unless explicitly excluded does not hold. It was always explicitly excluded unless the option "document.get_settings().use_revision" was enabled - I mean, you can simply look at the diff above.

I see the confusion you have with this. My take here is that the only "list"/set we need is the "set of excluded" fields. Same as it was before. You can default initialize the "exclude" set with the "revision_id" field, since we set that by default on the Document class. Then, if "document.get_settings().use_revision = True" we simply "remove" it from the set.
No need for a set of "included" items since there would really not be any point to it. What is the benefit of it and possible usage? As you mentioned, perhaps at a later stage we would altogether remove this Encoder class and handle it via Pydantic alone (maybe, one day).

beanie/odm/utils/encoder.py
staticxterm: It would be a big change to switch to model_dump(). Perhaps one day, at a later stage. Currently there is so much abstraction in Beanie (and the serialization logic is also all over the place) to do it right now.
However, pull requests are welcome.

beanie/odm/utils/dump.py
CatBraaain: > "But your reasoning that it was previously included unless explicitly excluded does not hold. It was always explicitly excluded unless the option `document.get_settings().use_revision` was enabled – I mean, you can simply look at the diff above."

Before this PR, it was entirely correct.

However, in this PR, the behavior of the `Encoder` class was changed so that `Field(exclude=True)` is *always* respected during serialization. As a result, the exclusion rules no longer behave the same as before. Unless we explicitly *include* the field (e.g., via `include={"revision_id"}`), it will be excluded regardless of the `exclude` argument.

To verify this behavior, I just pushed a new commit that adds a test case involving `use_revision`, and you can see that it fails. The failure is due to the fact that `revision_id` is not persisted during `insert()`—i.e., it is being excluded from serialization.

Below is a screenshot taken during debugging. Although `exclude={"revision_id"}` was not passed to the `Encoder`, the result of `_should_exclude_field()` is `True`, indicating that the field is still being excluded from serialization:




This confirms that the change in this PR causes `Field(exclude=True)` to be strictly honored inside the Encoder, which is a behavior change compared to before.

CatBraaain: @staticxterm
I updated this PR based on your suggestion, even though we had some discussion where we couldn't fully agree. After making the changes as you proposed, the tests are now failing.
This might indicate that the suggested change is breaking the intended behavior.
Could you please take another look when you have a moment? I'd really appreciate your feedback.

staticxterm: Hi @CatBraaain, okay, thank you, now I see what is going on here...
Previously, `revision_id` was always excluded unless the option "document.get_settings().use_revision" was enabled.
It was also not visible in output of e.g. `document.dict() / document.model_dump(mode="json")` since Pydantic respected the exclude option. Please note that Beanie internals for serialization do not use Pydantic methods but its own `get_dict()` method which serializes through the `Encoder` class.
Now, the only way to make it work using only the "exclude" set in the `Encoder` class is if we drop the `Field(..., exclude=True)` on the `revision_id` field. Since I guess this is not wanted, then I forfeit and please revert to your original implementation with the `include` set on the `Encoder` class as well - the problem is the check in `_should_exclude_field()`:
```python
return key in exclude or is_pydantic_excluded_field
```
Since it checks if either is valid it would always exclude it since it's defined to be excluded on the Field level.

beanie/odm/utils/encoder.py
staticxterm: Indeed, as you mentioned, due to changes in this PR, Pydantic's "Field(exclude=True)" in addition with Beanie's "exclude = {'revision_id'}" would always have priority and so it would always be excluded.
Previously it worked with only the "exclude" set since only that was what was respected when serializing through the Encoder class.
Resolving this discussion.

beanie/odm/utils/dump.py
CatBraaain: Thank you for your time and feedback. I've reverted to the original implementation.

staticxterm: Apologies for circling back to the original implementation – it wasn’t immediately clear to me why introducing a new include set was necessary.

Riverfount (approved):
LGTM

PR #1156, opened 2025-04-19
Keep Defaults settings

CAPITAINMARVEL:


staticxterm:
Hi @CAPITAINMARVEL,
the new functionality looks good. Is this Pydantic v2 related (e.g. the "warning" notice here: https://docs.pydantic.dev/latest/concepts/fields/#default-values) ?
I propose to add at least one test case showcasing this new argument (keep_defaults) and then also to update the docs with this new usage as well.

Issue #1161, opened 2025-04-25
[BUG] Naming of variable bucket_rounding_second(s) in TimeSeriesConfig


m00rex:
**Describe the bug**
https://github.com/BeanieODM/beanie/blob/main/beanie/odm/settings/timeseries.py
Code of TimeSeriesConfig uses the singular: "bucket_rounding_second"
https://beanie-odm.dev/tutorial/time-series/
Documentation of TimeSeriesConfig uses the plural: bucket_rounding_seconds

**To Reproduce**
```python
timeseries = TimeSeriesConfig(
            time_field="ts",
            meta_field="meta",
            granularity=Granularity.hours,
            bucket_max_span_seconds=3600,
            bucket_rounding_seconds=3600,
            expire_after_seconds=2
        )
```
gives Unexpected argument for "bucket_rounding_seconds=3600,"

**Expected behavior**
The code sample should not throw the warning

PR #1162, opened 2025-04-25
changed bucket_rounding_second to bucket_rounding_seconds in TimeSeri…

m00rex:
changed bucket_rounding_second to bucket_rounding_seconds in TimeSeries according to #1161


adeelsohailahmed (dismissed):
Good catch!

adeelsohailahmed:
This may break compatibility with the existing codebases that may have adapted to the (incorrect) `bucket_rounding_second` attribute. 

Could you please add that attribute back, but [set it as deprecated](https://docs.pydantic.dev/latest/concepts/fields/#deprecated-fields) so that the existing codebases may continue to work, if any?

adeelsohailahmed (approved):
Looks good to me. Thank you for the changes!

Issue #1106, opened 2025-01-06
Incorrect call arguments in PyCharm


MagicWinnie:
**Describe the bug**
When calling methods like `.save()`, `.insert()`, `.update()`, and `.delete()` I get the following warning from PyCharm:
```
Parameter(s) unfilled
Possible callees:(..., ...: [Any, Any])
```
Although, everything works fine, so it is something related to typing.

**To Reproduce**
```python
from beanie import Document


class Foo(Document):
    bar: int


async def func():
    foo = Foo(bar=1)

    await foo.save()  # warning here
    await foo.insert()  # warning here
    await foo.update()  # warning here
    await foo.delete()  # warning here
```

**Expected behavior**
No warning is displayed.

**Additional context**
PyCharm 2024.3.1.1  
Python 3.10
Beanie 1.28.0

dantetemplar:
upvote

MagicWinnie:
Also just putting
```python
AsyncDocMethod: TypeAlias = Callable[
    Concatenate[DocType, P], Coroutine[Any, Any, R]
]
```

in the top of `actions.py`  seems to silence the warnings

MagicWinnie:
Previous two solutions do not fix the issue

oneqit:
The decorator attached to those CRUD actions is:

```python
def wrap_with_actions(
    event_type: EventTypes,
) -> Callable[
    ["AsyncDocMethod[DocType, P, R]"], "AsyncDocMethod[DocType, P, R]"
]:
```
The return type of this function is `"AsyncDocMethod[DocType, P, R]"`, and when checking its definition:
```python
AsyncDocMethod: TypeAlias = Callable[
    Concatenate[DocType, P], Coroutine[Any, Any, R]
]
```
However, it seems that PyCharm recognizes it as Coroutine[Any, Any, R] instead.

hRtWzFe:
Created this PR as a quick possible solution. I really hope that there are better solutions than this

staticxterm:
Linking PyCharm issues related to this (there may be more):
https://youtrack.jetbrains.com/issue/PY-13276/Incorrect-call-arguments-inspection-doesnt-work-on-decorated-method
https://youtrack.jetbrains.com/issue/PY-48970/PyCharm-doesnt-report-mistakes-in-arguments-passed-to-a-decorated-function-call
https://youtrack.jetbrains.com/issue/PY-60104/Revise-the-way-we-inspect-decorated-declarations
[...]

PR #1166, opened 2025-04-28
Transform AsyncDocMethod[DocType, P, R] into Any to fix "Incorrect call arguments" warning in PyCharm

hRtWzFe:
As #1106 is still opened, creates much problems with projects and makes beanie almost unusable in PyCharm, and no other solutions satisfy both PyCharm type checker AND mypy type checker, replacement with "Any" looks like good idea

### DIFF
--- a/beanie/odm/actions.py
+++ b/beanie/odm/actions.py
@@ -206,7 +206,7 @@ def after_event(
 
 def wrap_with_actions(
     event_type: EventTypes,
-) -> Callable[[AsyncDocMethod[DocType, P, R]], Any]:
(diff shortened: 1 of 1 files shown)


staticxterm (approved):
I'm not opposed to having this PR merged, even though it "obfuscates" the real return type of the original method it "wraps", and may lead to other subtler bugs that may go unnoticed because of this.
E.g. the "decorator" inner-method below retains its return type typing which is now not reflected in the "wrap_with_actions" Callable return type.
All I can say is that there are more issues here with the "wrapper" as well due to the `#type: ignore` being used.

PR #1171, opened 2025-05-09
fix backlink serialization when link data is fetched

alexdlukens:
Hello,

I have noticed in FastAPI, if using a model with a BackLink as the return type for a route, the data will fail validation due to an error in validating the BackLink field.

On further inspection, it seems that there was no serialization handling for the BackLink when the field contains valid data from the linked model. 

I took the similar approach from the Link field, using the `serialize` method. Find the Link implementation [here](https://github.com/BeanieODM/beanie/blob/ea651add303c2e512eab54590ca6ff3542d55cd7/beanie/odm/fields.py#L388)

Added a test that checks for this to the FastAPI tests.

Best,
Alex

### DIFF
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -516,6 +516,11 @@ class Link(Generic[T]):
                 }
             )
 
+        def serialize(self):
+            if isinstance(self, Link):
+                return self.to_dict()
+            return self.dict()
+
     def to_ref(self):
         return self.ref
 
@@ -524,7 +529,7 @@ class Link(Generic[T]):
 
 
 if not IS_PYDANTIC_V2:
-    ENCODERS_BY_TYPE[Link] = lambda o: o.to_dict()
+    ENCODERS_BY_TYPE[Link] = lambda o: o.serialize()
 
 
 class BackLink(Generic[T]):
@@ -534,6 +539,12 @@ class BackLink(Generic[T]):
         self.document_class = document_class
 
     if IS_PYDANTIC_V2:
+
+        @staticmethod
+        def serialize(value: Union[BackLink[T], BaseModel]):
+            if isinstance(value, BackLink):
+                return value.to_dict()
+            return value.model_dump(mode="json")
 
         @classmethod
         def wrapped_validate(
@@ -565,7 +576,7 @@ class BackLink(Generic[T]):
                     values_schema=core_schema.any_schema(),
                 ),
                 serialization=core_schema.plain_serializer_function_ser_schema(
-                    lambda instance: cls.to_dict(instance),
+                    lambda instance: cls.serialize(instance),
                     return_schema=core_schema.dict_schema(),
                     when_used="json-unless-none",
                 ),
@@ -610,13 +621,18 @@ class BackLink(Generic[T]):
                 }
             )
 
+        def serialize(self):
+            if isinstance(self, BackLink):
+                return self.to_dict()
+            return self.dict()
+
     def to_dict(self) -> dict[str, str]:
         document_class = DocsRegistry.evaluate_fr(self.document_class)  # type: ignore
         return {"collection": document_class.get_collection_name()}
 
 
 if not IS_PYDANTIC_V2:
-    ENCODERS_BY_TYPE[BackLink] = lambda o: o.to_dict()
+    ENCODERS_BY_TYPE[BackLink] = lambda o: o.serialize()
 
 
 class IndexModelField:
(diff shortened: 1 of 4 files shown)


beanie/odm/fields.py
```diff
@@ -565,7 +576,7 @@ def __get_pydantic_core_schema__(
                     values_schema=core_schema.any_schema(),
                 ),
                 serialization=core_schema.plain_serializer_function_ser_schema(
-                    lambda instance: cls.to_dict(instance),
+                    lambda instance: cls.serialize(instance),
```
staticxterm, line 579: I've placed a comment here on line 569 noting to other developers that BackLinks should not be serialized (as I've understood from Roman's response in some GH issue comment).

tests/fastapi/routes.py
```diff
@@ -71,3 +71,8 @@ async def create_house_new(house: House = Body(...)):
     await house.save(link_rule=WriteRules.WRITE)
     await house.sync()
     return house
+
+
+@house_router.get("/person/{id}", response_model=Person)
```
staticxterm, line 76: This (response_model=Person) is not supported as BackLinks should not be serializable.
BackLink fetching (e.g. Person.get(fetch_links=True)) was never implemented nor specified as supported in the docs:
https://beanie-odm.dev/tutorial/relations/#back-links
> It is not possible to fetch() this virtual link after the initial search.

One problem being that it would lead to infinite recursion if "nesting_depth" was not specified, as you've done below.

tests/fastapi/test_api.py
```diff
@@ -66,3 +66,22 @@ async def test_create_house_new(api_client):
...
+
+    person_id = resp_json["owner"].get("id")
+    if person_id is None:
+        person_id = resp_json["owner"].get("_id")
+    assert person_id is not None
+
+    resp2 = await api_client.get(f"/v1/person/{person_id}")
+
+    resp2_json = resp2.json()
+    assert resp2_json["name"] == payload["owner"]["name"][-3:]
```
staticxterm, line 87: The assertion is kind of meaningless here. What is the point in fetching the related House "owner" from the document in a Person collection?
I struggle to see the real benefit of this functionality from this test alone.

beanie/odm/fields.py
```diff
@@ -535,6 +540,12 @@ def __init__(self, document_class: Type[T]):
 
     if IS_PYDANTIC_V2:
 
+        @staticmethod
+        def serialize(value: Union[BackLink[T], BaseModel]):
+            if isinstance(value, BackLink):
+                return value.to_dict()
+            return value.model_dump(mode="json")
```
staticxterm, line 547: What is the use case for this "else if not a BackLink" scenario? Please provide a test case showcasing this, similar  to the `test_id_types_serialized_when_dumping_to_json` test case which does it for the Link type.

staticxterm (changes requested):
Hi,
thank you for the PR.
Roman previously noted that BackLinks should only be "virtual" fields, and not be serializable.

Please provide a meaningful use case scenario where one could benefit from the serialization of a BackLink. Later, if there really is a valid use case for this, it should be added to the Beanie docs as well.

tests/fastapi/routes.py
alexdlukens: I am okay with not having the backlink field in the serialized result, but I believe serialization should not raise an error if the backlink field was fetched. Please see the minimal test I have added showing this behavior in [cf4a18d](https://github.com/BeanieODM/beanie/pull/1171/commits/cf4a18db7d1b93aa50068b6fd5cf54ce11f5de68)

tests/fastapi/test_api.py
alexdlukens: This example is a bit weak, but I think making CRUD routes for a database model is a common use case. As in routes that would insert, get, update, delete a database model. This flow is broken if we cannot serialize a model when a backlink has been fetched

I guess to make this work one would have to not include backlinks in the model at all, or never fetch_links...

alexdlukens:
Hello, please find the minimal test I have added to exhibit the problem I was seeing. This test raises an error when executed on the main branch:
```
FAILED tests/odm/test_relations.py::TestOther::test_dump_model_with_fetched_backlink - pydantic_core._pydantic_core.PydanticSerializationError: Error serializing to JSON: PydanticSerializationError: Error calling function `<lambda>`: AttributeError: 'DocumentWithLink' object has no attribute 'document_class'
```

alexdlukens:
Hello, I see the issue with the recursion as you mention. I still think the behavior I am seeing is a bug, but I understand why dumping the fetched model will not work as I have written.

GIven this, I am not sure what the expected result should be. If we want to return `{"collection": "<class_name>"}` in this situation, I propose something like in https://github.com/BeanieODM/beanie/compare/main...alexdlukens:beanie:fix-backlink-return-collection?expand=1

Best,

alexdlukens:
Superceded by PR #1207

Issue #874, opened 2024-02-21
[BUG] Migrations don't skip __init__.py file


thorin-schiffer:
**Describe the bug**
If migration directory has the init file in it, it will be executed as a migration file and marked in the migration history as a migration node.

**Expected behavior**
Skip the init file and not treat it as a migration.

roman-right:
Hi! Thank you for the catch. It is reasonable

PR #1177, opened 2025-05-19
Migration finder skips modules starting with underscore

unexceptable:
The new migration command always uses a timestamp at the start of the migration file name, while custom files can break that convention it is unlikely anyone will use _ as a starting character for migrations.

This largely is needed because otherwise the command thinks `__init__.py` is a migration file, and because of sorting, it ends up last.

Fixes: #874


staticxterm (approved):
Hi @unexceptable.
Thank you very much for this PR and congrats on your first contribution 🎉.
This has been on my to-do list of sorts to fix for a while now. 
> ...it is unlikely anyone will use _ as a starting character for migrations.

Yes, hopefully no one names their migrations with an underscore as a prefix.
With that said, the changes proposed here sound reasonable to me.

May I suggest that we also update the Beanie documentation with this information (and clarify about the migration naming):
https://beanie-odm.dev/tutorial/migrations/
(here is the file to update: https://github.com/BeanieODM/beanie/blob/main/docs/tutorial/migrations.md)
It could be in a separate PR, although there may be more "friction" to merge it.

unexceptable:
@staticxterm Thanks! You'll probably see more misc fixes from me as I get frustrated with things enough to fix them, or at least document them better :P

I'll update this PR with docs, as it would be saner to just do the two together imo.

staticxterm (approved):
Looks great! Thank you very much.

adeelsohailahmed (approved):
Thanks for your contribution!

Issue #1057, opened 2024-10-17
[BUG] update() raises a `RevisionIdWasChanged` when expection a `DuplicateKeyError`


NiziL:
**Describe the bug**

When creating a `Document` with an unique indexed attribute, violating the uniqueness raise a `RevisionIdWasChanged` instead of a `DuplicateKeyError`.

**To Reproduce**
```python
class MyDocument(Document):
    unique: Annotated[str, Indexed(index_type=pymongo.TEXT, unique=True)]

... # beanie init

doc = MyDocument(unique="test")
doc.save()
doc2 = MyDocument(unique="test")
doc2.save()
```

**Expected behavior**
Raise the `DuplicateKeyError` instead of replacing it by a `RevisionIdWasChanged`.

**Additional context**
The bug comes from [documents.py, line 739/740](https://github.com/BeanieODM/beanie/blob/main/beanie/odm/documents.py#L739).

I feel like we could easily catch the error, ensure the `DuplicateKeyError` comes from a revision id issue, and either raise a `RevisionIdWasChanged` or this error.  
Something like:
```python
except DuplicateKeyError as e:
    revision_issue = ...
    raise RevisionIdWasChanged if revision_issue else e 
```

I can prepare a PR for this, I might just need some help to properly find the case where the revision id is the true issue.  
Maybe someone can just tell me where is the test for this case, so I can ensure I properly raise the `RevisionIdWasChanged`.

RamonGiovane:
Probably another side effect of this bug: https://github.com/BeanieODM/beanie/issues/833#issuecomment-2113163748

mikeckennedy:
I'm getting this error when inserting a new document with no revision set at all but there is a composite key that is set to unique.

martincpt:
Can't we just compare if really the `revesion_id` was what changed or another field? 

I agreee with OPs solution. Maybe it's not bulletproof but works out of the box.

nagarajRPoojari:
```python
if use_revision_id:
    new_revision_id = uuid4()
    arguments.append(SetRevisionId(new_revision_id))

if use_revision_id and not ignore_revision:
    fetch_result = await self.find_one(find_query)
    if fetch_result and fetch_result.revision_id != self.revision_id:
        raise RevisionIdWasChanged

try:
    result = await self.find_one(find_query).update(
        *arguments,
        session=session,
        response_type=UpdateResponse.NEW_DOCUMENT,
        bulk_writer=bulk_writer,
[...]

PR #1180, opened 2025-05-31
fix: raise DuplicateKeyError instead of RevisionIdWasChanged on duplicate insert

nagarajRPoojari:
issue:  #1057 

fix: 
- add DuplicateKeyError to beanie exceptions
- raise DuplicateKeyError instead of RevisionIdWasChanged on duplicate insert


martincpt:
I think you should check at least the error message to see what field is affected.

See #1057 OPs suggestion.

`RevisionIdWasChanged` is there for a good reason.

Issue #1182, opened 2025-06-07
[BUG] fetch_links


Mahdi-Ba:
I write a code like below, and each Conversation has link to other models like Tag, but when filtering based on more than condition, fetch_links=True changes to False 
```python
 query = Conversation.find(Conversation.user == user_id,fetch_links=True)
 if updated_at_from is not None:
      query = query.find(Conversation.updated_at >= updated_at_from)
```
Therefore, I rewrote it like this 

```python
 query = Conversation.find(Conversation.user == user_id,fetch_links=True)
 if updated_at_from is not None:
      query = query.find(Conversation.updated_at >= updated_at_from,fetch_links=True)
```

PR #1184, opened 2025-06-09
fix: preserve fetch_links across chained find() and find_one()

scarlet2131:
### What

This PR fixes a bug where `fetch_links=True` was not preserved when chaining multiple `.find()` or `.find_one()` calls.

### Why

As reported in [issue #1182](https://github.com/BeanieODM/beanie/issues/1182), chaining `.find()` caused the `fetch_links` flag to be dropped, resulting in unresolved linked documents unless manually re-added.

### Fix

This change ensures that `fetch_links`, `nesting_depth`, and related flags are retained across chained queries by setting them explicitly during `.find()` reuse.

### Test

Reproduced the issue locally using a minimal example with linked models and validated that the bug no longer occurs after this fix.

Before the fix : 

Screenshot 2025-06-09 at 2 05 41 PM

After the fix : 

Screenshot 2025-06-09 at 2 05 00 PM

### DIFF
--- a/tests/odm/query/test_find.py
+++ b/tests/odm/query/test_find.py
@@ -8,11 +8,11 @@ from beanie.odm.enums import SortDirection
 from tests.odm.models import (
     Color,
     DocumentWithBsonEncodersFiledsTypes,
+    Door,
     House,
+    Lock,
     Sample,
     Window,
-    Door,
-    Lock
 )
 
 
@@ -436,10 +436,12 @@ async def test_fetch_links_with_chained_delete():
     await House(windows=[window], door=door, height=10, name="test").insert()
     await House(windows=[window], door=door, height=12, name="test2").insert()
 
-   # Deletion with chained query and fetch_links
-    deleted_count = await House.find(
-        House.height > 5, fetch_links=True
-    ).find(House.height < 20).delete()
+    # Deletion with chained query and fetch_links
+    deleted_count = (
+        await House.find(House.height > 5, fetch_links=True)
+        .find(House.height < 20)
+        .delete()
+    )
 
     assert deleted_count.deleted_count == 2
 
--- a/tests/odm/test_relations.py
+++ b/tests/odm/test_relations.py
@@ -438,28 +438,37 @@ class TestFind:
         door = await Door(t=10, window=window, locks=[lock]).insert()
 
         # Inserted the houses to delete
-        await House(windows=[window], door=door, height=10, name="test").insert()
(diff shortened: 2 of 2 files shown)


scarlet2131:
@roman-right @adeelsohailahmed  sir, I’ve verified the issue and submitted a fix in this PR 
The CL preserves fetch_links across chained find() calls by carrying over the flag from the original query instance.

Tests pass locally 
GitHub Actions failed due to Docker rate limits when pulling the mongo:7.0 image, not related to the code change.

Let me know if anything needs to be updated!

staticxterm:
Hi @scarlet2131,
thank you very much for the PR.
Could you please mention the usage and this behavior in the Beanie documentation pages as well, for the find() examples:
https://beanie-odm.dev/tutorial/finding-documents/
https://github.com/BeanieODM/beanie/blob/main/docs/tutorial/find.md

What is worrying me is this "limitation" description as mentioned here:
https://beanie-odm.dev/tutorial/relations/#limitations
Could you please write some tests for the chaining examples to have it covered with tests. Also, to see what error is raised when chaining e.g. `find(..., fetch_links=True).delete()` as mentioned previously.

adeelsohailahmed:
Could you please add an automated test for this use case too, @scarlet2131?

[This file](https://github.com/BeanieODM/beanie/blob/f6dc433b27488be27f50a8d79dc413ce9b256408/tests/odm/test_relations.py) contains several tests for the Links. I suppose you could extend one of these classes and add a test there.

scarlet2131:
sure @adeelsohailahmed @staticxterm @CAPITAINMARVEL , working on the tests will ping here once added

scarlet2131:
Hi @adeelsohailahmed, @staticxterm, and @CAPITAINMARVEL,

I have added tests to confirm that using fetch_links=True with chained .find() now works correctly when followed by .delete() and .update() operations.

These tests are added in:

test_find.py for basic chaining and deletion

test_relations.py to verify behavior with linked documents and to show that deletion and update now work as expected

The tests cover:

Chaining multiple .find() calls with fetch_links=True

Deleting and updating documents in those chained queries

Ensuring that the linked documents are correctly handled

I also updated the documentation in docs/tutorial/find.md to show an example of how this chaining works with fetch_links.

All tests are passing locally. Please let me know if there's anything else I should do.

Mahdi-Ba:
pleaase merge it

staticxterm (approved):

PR #1188, opened 2025-06-14
Added SoftDelete Event

### LINKED ISSUE #1187: [BUG] `DocumentWithSoftDelete.delete` does not trigger `DELETE` event
**Describe the bug**
The `DocumentWithSoftDelete.delete` is not wrapped with the `DELETE` event and therefore does not trigger the expected event.

**To Reproduce**
```python
class Foo(DocumentWithSoftDelete):
    foo: str = 'bar'

    @before_event(Delete)
    async def on_before_delete():
        # NEVER GETS CALLED
        await do_other_things()

await Foo.find_one(...).delete()
```

**Expected behavior**
`DocumentWithSoftDelete.delete` should invoke the `DELETE` document event.

kaperont:
Resolves #1187 

Added a `SoftDelete` event type and hooked it up to `DocumentWithSoftDelete.delete()`.


beanie/odm/documents.py
```diff
@@ -1283,7 +1284,9 @@ async def delete(
         **pymongo_kwargs,
     ) -> Optional[DeleteResult]:
         self.deleted_at = datetime.now(tz=timezone.utc)
-        await self.save()
+        await self.save(
+            skip_actions=[ActionDirections.BEFORE, ActionDirections.AFTER]
+        )
```
kaperont, line 1289: Not sure if skipping actions here is correct, but it felt semantically correct since this is an "internal action" and I think it's better not to trigger any potentially defined side-effects of Save/Insert/Update.

kaperont:
Any update on this?

PR #1190, opened 2025-06-15
feat: Added support for serialization_alias in beanie

### LINKED ISSUE #1112: [BUG]: Serialization_alias on pydantic
**Describe the bug**
Beanie do not use the `serialization_alias` using `Pydantic`.

**To Reproduce**
```python
class A(Dcoument, BaseModel):
    foo: str = Field(serialization_alias="bar")

a = A(foo="zoo")

a.save()
```

**Expected behavior**
The behavior should be exactly the same as if I was using the `alias` parameter in the Field.
When saving the document I want to see the aliased name.

**Additional context**

abdul-haseeb123:
### Describe the problem
Beanie do not use the `serialization_alias` using Pydantic.

### To Reproduce
```python
class A(Dcoument, BaseModel):
    foo: str = Field(serialization_alias="bar")

a = A(foo="zoo")
a.save()
```
###Expected behavior
The behavior should be exactly the same as if using the alias parameter in the Field.
The document should be using the aliased name when saving the document

### Solution
`beanie/odm/utils/init.py`
```diff
    def init_document_fields(self, cls) -> None:
        """
        Init class fields
        :return: None
        """

        if not IS_PYDANTIC_V2:
            self.update_forward_refs(cls)

        if cls._link_fields is None:
            cls._link_fields = {}
        for k, v in get_model_fields(cls).items():
-            path = v.alias or k
+            path = v.serialization_alias or v.alias or k
            setattr(cls, k, ExpressionField(path))
```

serialization_alias does now gets priority when encoding the data.
`beanie/odm/utils/encoder.py`
```diff
    def _iter_model_items(
        self, obj: pydantic.BaseModel
    ) -> Iterable[Tuple[str, Any]]:
        exclude, keep_nulls = self.exclude, self.keep_nulls
        get_model_field = get_model_fields(obj).get
        for key, value in obj.__iter__():
            field_info = get_model_field(key)
            if field_info is not None:
-                key = field.alias or key
+                key = field_info.serialization_alias or field_info.alias or key
            if key not in exclude and (value is not None or keep_nulls):
                yield key, value
```
Additional conversion of fields that are serialized required before model validation.
`beanie/odm/utils/pydantic.py`
```diff
def parse_model(model_type: Type[BaseModel], data: Any):
+    for k, field in get_model_fields(model_type).items():
+       if field.alias and field.alias != field.serialization_alias:
+           data[k] = data[field.serialization_alias]
+            del data[field.serialization_alias]
    if IS_PYDANTIC_V2:
        return model_type.model_validate(data)
    else:
        return model_type.parse_obj(data)
```

### DIFF
--- a/beanie/odm/utils/init.py
+++ b/beanie/odm/utils/init.py
@@ -516,7 +516,9 @@ class Initializer:
                 IndexModel(
                     [
                         (
-                            fvalue.alias or k,
+                            getattr(fvalue, "serialization_alias", None)
+                            or fvalue.alias
+                            or k,
                             indexed_attrs[0],
                         )
                     ],
@@ -639,7 +641,7 @@ class Initializer:
         if cls._link_fields is None:
             cls._link_fields = {}
         for k, v in get_model_fields(cls).items():
-            path = v.alias or k
+            path = getattr(v, "serialization_alias", None) or v.alias or k
             setattr(cls, k, ExpressionField(path))
             link_info = self.detect_link(v, k)
             depth_level = cls.get_settings().max_nesting_depths_per_field.get(
--- a/beanie/odm/utils/projection.py
+++ b/beanie/odm/utils/projection.py
@@ -32,5 +32,7 @@ def get_projection(
     document_projection: Dict[str, int] = {}
 
     for name, field in get_model_fields(model).items():
-        document_projection[field.alias or name] = 1
+        document_projection[
+            getattr(field, "serialization_alias", None) or field.alias or name
+        ] = 1
     return document_projection
--- a/tests/odm/test_beanie_serialization.py
+++ b/tests/odm/test_beanie_serialization.py
@@ -20,7 +20,7 @@ async def test_serialization_types_preserved_after_insertion():
     assert document is not None
     assert document.test_int is not None
     assert document.test_str is not None
-    dumped = document.model_dump()
+    dumped = document.model_dump(by_alias=True)
     assert "test_int_serialize" in dumped
     assert "test_str_serialize" in dumped
     assert "test_int" not in dumped


beanie/odm/utils/encoder.py
```diff
@@ -159,7 +159,7 @@ def _iter_model_items(
         for key, value in obj.__iter__():
             field_info = get_model_field(key)
             if field_info is not None:
-                key = field_info.alias or key
+                key = field_info.serialization_alias or field_info.alias or key
```
staticxterm: For Pydantic v1, this field does not exist so you would have to use `field_info.get("serialization_alias")`

beanie/odm/utils/init.py
```diff
@@ -403,7 +403,7 @@ def init_document_fields(self, cls) -> None:
         if cls._link_fields is None:
             cls._link_fields = {}
         for k, v in get_model_fields(cls).items():
-            path = v.alias or k
+            path = v.serialization_alias or v.alias or k
```
staticxterm: Same comment as for "_iter_model_items".

beanie/odm/utils/pydantic.py
```diff
@@ -36,6 +36,10 @@ def get_model_fields(model):
 
 
 def parse_model(model_type: Type[BaseModel], data: Any):
+    for k, field in get_model_fields(model_type).items():
+        if field.alias and field.alias != field.serialization_alias:
+            data[k] = data[field.serialization_alias]
+            del data[field.serialization_alias]
```
staticxterm: This would have to be reworked. I would just add the whole loop under "if IS_PYDANTIC_V2" check, preserving the "old" code and behavior for Pydantic v1 in the "else" branch.

staticxterm (dismissed):
Looks great, thank you for the PR!

adeelsohailahmed (changes requested):
Add a test

abdul-haseeb123:
@adeelsohailahmed I have a good understanding of what tests need to be written, but could you please guide me on where exactly they should be placed within the `tests` directory?

abdul-haseeb123:
@adeelsohailahmed @staticxterm I have added test, but I think that they aren't doing enough coverage, and will be adding some more.
### What has been implemented
Projection model is now using the serialized fields for retrieving the data
```diff
def get_projection(
    model: Type[ProjectionModelType],
) -> Optional[Dict[str, int]]:
    if hasattr(model, "get_model_type") and (
        model.get_model_type() == ModelType.UnionDoc  # type: ignore
        or (  # type: ignore
            model.get_model_type() == ModelType.Document  # type: ignore
            and model._inheritance_inited  # type: ignore
        )
    ):  # type: ignore
        return None

    if hasattr(model, "Settings"):  # MyPy checks
        settings = getattr(model, "Settings")

        if hasattr(settings, "projection"):
            return getattr(settings, "projection")

    if get_config_value(model, "extra") == "allow":
        return None

    document_projection: Dict[str, int] = {}

    for name, field in get_model_fields(model).items():
-        document_projection[field.alias or name] = 1
+        document_projection[
+           getattr(field, "serialization_alias", None) or field.alias or name
+        ] = 1
    return document_projection
 ```
 Also when initializing indexes, serialization_alias is now getting the priority over the alias field.
**Note that the alias field is added to all user defined field annotations in the document by default from LazyModel**
so it doesn't matter whether you set the alias field or not, it is always automatically set.

here are the code changes when initializing index fields
```diff
found_indexes = [
            IndexModelField(
                IndexModel(
                    [
                        (
-                          fvalue.alias or k
+                          getattr(fvalue, "serialization_alias", None)
+                          or fvalue.alias
+                         or k,
                            indexed_attrs[0],
                        )
                    ],
                    **indexed_attrs[1],
                )
            )
            for k, fvalue, indexed_attrs in indexed_fields
            if indexed_attrs is not None
        ]
```

so now the indexes are also initialized using serialization alias and I have also checked it with mongodb, but couldn't write a test for it as I don't know how to access indexes of a model, I had tried to use `document._document_settings.indexes`, but it provided an empty list.


**Note that although I have added the same logic for `init_view_fields` as well, but it breaks the code inside the `parse_model()`, reason for that is `View` class uses `BaseModel` instead of `LazyModel`, which does not set any `alias` automatically to any field and it then breaks this condition inside `parse_model()`
`if  field.alias and field.alias != field.serialization_alias` as now there is no alias by default on any field annotation.

beanie/odm/utils/init.py
```diff
@@ -516,7 +516,9 @@ async def init_indexes(self, cls, allow_index_dropping: bool = False):
                 IndexModel(
                     [
                         (
-                            fvalue.alias or k,
+                            getattr(fvalue, "serialization_alias", None)
```
staticxterm, line 519: Likewise here, this would be a breaking change if the user does not have the `allow_index_dropping=True` flag set to recreate the indexes with the new preferred names.

Please add some more tests here around indexes and for these scenarios, e.g. model with previous indexes without the serialization alias, and another model (copy of the previous one) that has the `serialization_alias`es set, check what happens with indexes on the DB side...

beanie/odm/utils/init.py
```diff
@@ -403,7 +403,7 @@ def init_document_fields(self, cls) -> None:
         if cls._link_fields is None:
             cls._link_fields = {}
         for k, v in get_model_fields(cls).items():
-            path = v.alias or k
+            path = getattr(v, "serialization_alias", None) or v.alias or k
             setattr(cls, k, ExpressionField(path))
```
staticxterm, line 407: _Somehow this comment got lost..._
This would be a breaking change, and probably would not be correct if someone used the `validation_alias` instead to initialize the model...

staticxterm (changes requested):

beanie/odm/utils/init.py
abdul-haseeb123: got it, thanks for the detailed overview, will look into it.

abdul-haseeb123:
@staticxterm 
> **Note that although I have added the same logic for `init_view_fields` as well, but it breaks the code inside the `parse_model()`, reason for that is `View` class uses `BaseModel` instead of `LazyModel`, which does not set any `alias` automatically to any field and it then breaks this condition inside `parse_model()` `if field.alias and field.alias != field.serialization_alias` as now there is no alias by default on any field annotation.**

what is your thought on this ?

staticxterm:
@abdul-haseeb123,
I don't have any standpoint on this since I haven't ever used View's in Beanie myself.
I would have to look more deeply into this before I can comment on that, to understand how it was meant to be used and how it should behave.
Perhaps if you can come up with some test with an usage example, and if it can also demonstrate this breakage that would be great.
Thanks

beanie/odm/utils/init.py
abdul-haseeb123: @staticxterm yeah, you were right, I first created a model with a field having index, but no serialization_alias, and it created the correct index, then I modified the field to have a serialization_alias, and it also created a index according to the serialization_alias but doesn't drop the previous one, for it to gets drop, I tried allow_index_dropping=True and then it drops that index.

staticxterm:
Hi @abdul-haseeb123, are you still working on this?
I have added similar support in parallel, supporting also the AliasGenerator at the same time.
I was planning on pushing the changes to this PR to support both.

abdul-haseeb123:
> Hi @abdul-haseeb123, are you still working on this? I have added similar support in parallel, supporting also the AliasGenerator at the same time. I was planning on pushing the changes to this PR to support both.

I was working on it, a month or when I was active on this issue and I has also found the solution and had also written some tests for it too,


Write now on my phone, I will give you a detailed response tomorrow

staticxterm:
Hi @abdul-haseeb123, any update on this?
I plan to push some changes here, to add better support for alias_generator as well. So both serialization and deserialization from DB would be covered.

ranfysvalle02:
@staticxterm -- https://github.com/abdul-haseeb123/beanie/pull/1/files -> here is an attempt to revive a stale PR. 

Do you prefer if I start a new PR/branch based on this PR? 

@abdul-haseeb123 Anything I can do to help get this down the line?

roman-right:
This has been open for a while with merge conflicts and outstanding change requests. The feature itself (serialization_alias support) is valid, but this PR needs a fresh start. If someone wants to pick this up, please open a new PR with a clean rebase and address the review comments about tests and the breaking change with alias priority. Closing for now.

PR #1192, opened 2025-06-15
Fix PytestCollectionWarning in tests

staticxterm:
Fixes the PytestCollectionWarning warning which becomes visible once we remove suppression of all warnings.


adeelsohailahmed (commented):
How about this instead?

```python
class TestModel(BaseModel):
    my_id: PydanticObjectId = Field(default_factory=PydanticObjectId)
    fake_doc: Link[DocumentTestModelWithSoftDelete]

    __test__: bool = False
```

staticxterm:
Yeah @adeelsohailahmed, this is valid as well but it's one more line to add/change, and we shouldn't really name things starting with "Test" due to this handling in pytest.

adeelsohailahmed:
One line change vs two line changes. Take your pick.

staticxterm:
I would like to discourage this "naming convention" (if it can even be called that), so I've opted for the rename instead.

adeelsohailahmed:
Eh, it's just one model. That hardly constitutes as "convention".

PR #1158, opened 2025-04-21
feat: support `@computed_field` of pydantic v2

CatBraaain:
This PR adds support for `@computed_field` to Beanie, which is currently not supported.

reference:
- [model_fields](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_fields)
- [model_computed_fields](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_computed_fields)

PR #1200, opened 2025-07-11
Update encoder.py

mg3146:
The current encoder method iterates through obj.iter. However this captures class properties that are not explicitly defined in the base model as a field (or as a computed field).

This can cause bugs where a user has a property that is not encodable [and was not meant to be a field in the Document].

I limited it to only used keys that are set as fields or computed fields.

### DIFF
--- a/beanie/odm/utils/encoder.py
+++ b/beanie/odm/utils/encoder.py
@@ -151,27 +151,27 @@ class Encoder:
 
         raise ValueError(f"Cannot encode {obj!r}")
 
-    def _iter_model_items(
-        self, obj: pydantic.BaseModel
-    ) -> Iterable[Tuple[str, Any]]:
+    def _iter_model_items(self, obj: pydantic.BaseModel) -> Iterable[Tuple[str, Any]]:
         exclude, keep_nulls = self.exclude, self.keep_nulls
         get_model_field = get_model_fields(obj).get
 
-        obj_iter_keys = obj.__iter__.keys()
         if IS_PYDANTIC_V2:
-            if obj.__class__.model_config.get("extra") != "allow":
-                model_class = obj.__class__
-                obj_iter_keys = list(model_class.model_fields.keys()) + list(
-                    model_class.model_computed_fields.keys()
-                )
-
+            model_class = obj.__class__``
+            if model_class.model_config.get("extra") != "allow":
+                obj_iter_keys = set(model_class.model_fields.keys()) | set(
+                model_class.model_computed_fields.keys()
+            )
+            else:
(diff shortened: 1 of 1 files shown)


mg3146:
I should have said this was for the case of encoding a nested base model

mg3146 (commented):
fix unbounded error

mg3146 (commented):
allow extras

mg3146 (commented):
fixed iter error

staticxterm:
Hi @mg3146 ,
have you checked the other open PR that also introduces this same support: #1158 ? I feel like this is a duplicate of that.
As you noticed, we have to iterate through all items (even non-existing fields) due to 'extras' support.

mg3146:
Very similar. My approach I guess inadvertently addresses that by using a more pydanitc-friendly key lookup in that encoder loop.   However I dont think that addresses the case where you have a Document with a nested BaseClass that contains a property with a non-encodable type. 

This is probably a holdover from the Pydantic 1 days - but it seems like there probably a number of places where the move to pure pydantic 2 will makes things a lot easier.

staticxterm:
Hi @mg3146,

honestly, you still keep the `for key, value in obj.__iter__():` old-style loop, so nothing is really gained from this. If anything, it complicates the code.
In my opinion, the other PR that adds support for "Pydantic computed fields" does this better since it extracts that "field gathering and handling" complexity away to a helper method.
The only reason I comment on this part of code is that I was working recently on it as well, trying to add better support for aliases (serialization_alias, AliasGenerator) in Pydantic v2.

I do agree that once we move to Pydantic v2 it may be easier to do some changes. But first we should work towards better support for Pydantic v2 (after we remove support for Pydantic v1, to keep everything simpler).

mg3146:
I hear where you're coming from for the computed_fields, however in my case my original error wasn't from a computed_field.  It was something more like this [I dont really have an easy way to test if this causes my error.. but I think it will assuming that np.array can't be serialized].

Totally makes sense the serialization alias stuff too.  I was thinking about that, how it would seem that a lot of that is missed.  I wans't sure if things like custom serializers defined on the pydantic class would get picked up or not


```
from pydantic import BaseModel
from functools import cached_property
from beanie import Document
import numpy as np

class A(Document):
    field_1 : str
    field_2 : int
    b : B
    
class B(BaseModel):
    x : int
    y : float
    
    @cached_property
    def z(self):
        return np.array([self.x, self.y])
    

b = B(x = 10, y = 25.5)
print(b.z)  #<--- this will put the output of 'z' into the instances __dict__, and so gets picked up by the iter loop.

a = A(field_1 = "abc", field_2 = 123, b = b)
await a.insert()

```

PR #1203, opened 2025-07-11
handle aggregation method on the whole collection

CAPITAINMARVEL:
support aggregation method specified in the beanie document (shortcut without needing .find_all()
https://beanie-odm.dev/tutorial/aggregation/

### DIFF
--- a/tests/odm/query/test_aggregate_methods.py
+++ b/tests/odm/query/test_aggregate_methods.py
@@ -98,11 +98,11 @@ async def test_min_without_docs(session):
 async def test_all_sum(preset_documents, session):
     n = await Sample.sum(Sample.increment)
 
-    assert n == 12
+    assert n == 45
 
     n = await Sample.sum(Sample.increment, session=session)
 
-    assert n == 12
+    assert n == 45
 
 
 async def test_all_sum_without_docs(session):
@@ -118,10 +118,10 @@ async def test_all_sum_without_docs(session):
 async def test_all_avg(preset_documents, session):
     n = await Sample.avg(Sample.increment)
 
-    assert n == 4
+    assert n == 4.5
     n = await Sample.avg(Sample.increment, session=session)
 
-    assert n == 4
+    assert n == 4.5
 
 
 async def test_all_avg_without_docs(session):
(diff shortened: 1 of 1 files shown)


beanie/odm/interfaces/aggregate.py
```diff
@@ -68,3 +69,39 @@ def aggregate(
             ignore_cache=ignore_cache,
             **pymongo_kwargs,
         )
+
+    @classmethod
+    async def sum(
```
staticxterm, line 74: Please add docstrings to all the new public API methods, similarly as for the above `aggregate` method.
Otherwise the docstring would not be visible in e.g. VSCode when hovering over "avg" for the second example from Beanie documentation: `avg_price = await Product.avg(Product.price)`.

beanie/odm/interfaces/aggregate.py
```diff
@@ -68,3 +69,39 @@ def aggregate(
...
+        field: Union[str, ExpressionField],
+        session: Optional[AsyncClientSession] = None,
+        ignore_cache: bool = False,
+    ) -> Optional[float]:
+        return await cls.find_all().sum(field, session, ignore_cache)
+
+    @classmethod
+    async def avg(
+        cls,
+        field,
```
staticxterm: Missing types.
```suggestion
        field: Union[str, ExpressionField],
```

beanie/odm/interfaces/aggregate.py
staticxterm, line 74: Please also add tests that test or showcase the usage of these new methods, e.g.  in `tests/odm/query/test_aggregate_methods.py` or wherever appropriate.

beanie/odm/interfaces/aggregate.py
CAPITAINMARVEL: min and max can be any (so i gave same type hint as the set method)
avg sum can only be a numeric value (mongodb ignore non numeric) so should be fixed

beanie/odm/interfaces/aggregate.py
CAPITAINMARVEL: docstrings added

adeelsohailahmed (approved):

PR #1204, opened 2025-07-15
Fix return types of min and max methods

kewldan:
Hello, aggregations "min" and "max" can also return types other than float, such as: int, datetime. I changed the return types to using Union for backward compatibility. Also the pydoc for the methods was changed.

Make Beanie [fully type annotated](https://github.com/BeanieODM/beanie/issues/1050)!

### DIFF
--- a/beanie/odm/interfaces/aggregation_methods.py
+++ b/beanie/odm/interfaces/aggregation_methods.py
@@ -22,7 +22,7 @@ class AggregateMethods:
 
     async def sum(
         self,
-        field: Union[str, ExpressionField],
+        field: Union[ExpressionField, float, int, str],
         session: Optional[AsyncClientSession] = None,
         ignore_cache: bool = False,
     ) -> Optional[float]:
@@ -41,7 +41,7 @@ class AggregateMethods:
 
         ```
 
-        :param field: Union[str, ExpressionField]
+        :param field: Union[ExpressionField, float, int, str]
         :param session: Optional[AsyncClientSession] - pymongo session
         :param ignore_cache: bool
         :return: float - sum. None if there are no items.
@@ -66,7 +66,7 @@ class AggregateMethods:
 
     async def avg(
         self,
-        field,
+        field: Union[ExpressionField, float, int, str],
         session: Optional[AsyncClientSession] = None,
         ignore_cache: bool = False,
     ) -> Optional[float]:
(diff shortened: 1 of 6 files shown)


adeelsohailahmed (dismissed):
Thanks, Dan. Very cool!

beanie/odm/interfaces/aggregation_methods.py
```diff
@@ -152,7 +153,7 @@ async def min(
         field: Union[str, ExpressionField],
         session: Optional[AsyncClientSession] = None,
         ignore_cache: bool = False,
-    ) -> Optional[float]:
+    ) -> Optional[Union[int, float, datetime]]:
```
staticxterm:
```suggestion
    ) -> Optional[Any]:
```

beanie/odm/interfaces/aggregation_methods.py
```diff
@@ -111,7 +112,7 @@ async def max(
...
```
staticxterm:
```suggestion
    ) -> Optional[Any]:
```
My 2 cents are that it's best to annotate as `Any` since we can't control or know beforehand which type it would be.

beanie/odm/interfaces/aggregation_methods.py
```diff
@@ -128,7 +129,7 @@ class Sample(Document):
 
         :param field: Union[str, ExpressionField]
         :param session: Optional[AsyncClientSession] - pymongo session
-        :return: float - max. None if there are no items.
+        :return: int, float or datetime - max. None if there are no items.
```
staticxterm:
```suggestion
        :return: Any - max value. None if there are no items.
```

beanie/odm/interfaces/aggregation_methods.py
```diff
@@ -169,7 +170,7 @@ class Sample(Document):
 
         :param field: Union[str, ExpressionField]
         :param session: Optional[AsyncClientSession] - pymongo session
-        :return: float - min. None if there are no items.
+        :return: int, float or datetime - min. None if there are no items.
```
staticxterm:
```suggestion
        :return: Any - min value. None if there are no items.
```

adeelsohailahmed:
> My recommendation is to use `Any` instead of the few mentioned types which would be "the usual suspects" (common return types) since we can't control this and there may be more types that the user can play with (e.g. Binary).
> 

I disagree with this approach.

These three possible values, along with the `None` (from `Optional`) are more than enough. Can you think of any other type that would be returned by a possible min or max?

staticxterm:
@adeelsohailahmed  Yes, I can think of many, and @CAPITAINMARVEL  already posted a screen of this in the Beanie Discord #contributors private channel.
E.g. bytes (from the BinData Mongo type / Binary BSON type), UUID, Timestamp, Decimal128.
Just because something is more reasonable than the rest doesn't mean we should limit it only to that.
So I also respectfully disagree and my proposal is to not type check this at all (use the `Any` type).

Riverfount (dismissed):
LGTM

staticxterm (approved):
Looks good to me too. Thank you!

Riverfount (approved):
LGTM

Issue #1189, opened 2025-06-14
[BUG] Model fails to serialize when BackLink has been fetched


alexdlukens:
Hello all,

**Describe the bug**
When a model with a BackLink has been retrieved (with `fetch_links=1, nesting_depth=1`), calling `model_dump` or `model_dump_json` raises a `PydanticSerializationError`

**To Reproduce**
```python
from beanie import init_beanie, Document, BackLink, Link
from motor.motor_asyncio import AsyncIOMotorClient
from pydantic import Field

class User(Document):
    name: str
    login: Link["Login"]

class Login(Document):
    username: str
    password: str
    user: BackLink[User] = Field(json_schema_extra={"original_field": "login"})
    
User.model_rebuild()

client = AsyncIOMotorClient("mongodb://localhost:27017")
db = client.get_database("test_db")


await init_beanie(
    db,
    document_models=[User, Login],
)

login = Login(username="alice123", password="securepassword")
await login.save()
user1 = User(name="Alice", login=login)
await user1.save()

login_doc = await Login.get(login.id, fetch_links=True, nesting_depth=1)
login_doc.model_dump_json() # error is raised here
```
**Expected behavior**

No exception should be thrown. Serialization should succeed, with or without the backlinked data

**Additional context**

This was the issue I attempted to solve in #1171, but my resolution was not approved. Just looking for any solution to fix the serialization issue

alexdlukens:
here is the traceback:
```
---------------------------------------------------------------------------
PydanticSerializationError                Traceback (most recent call last)
Cell In[1], line 31
     28 await user1.save()
     30 login_doc = await Login.get(login.id, fetch_links=True, nesting_depth=1)
---> 31 login_doc.model_dump_json() # error is raised here

File .venv/lib/python3.12/site-packages/pydantic/main.py:519, in BaseModel.model_dump_json(self, indent, include, exclude, context, by_alias, exclude_unset, exclude_defaults, exclude_none, round_trip, warnings, fallback, serialize_as_any)
    479 def model_dump_json(
    480     self,
    481     *,
   (...)    493     serialize_as_any: bool = False,
    494 ) -> str:
    495     """!!! abstract "Usage Documentation"
    496         [`model_dump_json`](../concepts/serialization.md#modelmodel_dump_json)
    497 
   (...)    517         A JSON string representation of the model.
    518     """
--> 519     return self.__pydantic_serializer__.to_json(
    520         self,
    521         indent=indent,
    522         include=include,
[...]

PR #1207, opened 2025-07-19
use class from self.__class__ when serializing BackLinks with fetched data

alexdlukens:
This is my proposed solution for #1189. Please find the minimal code example in the issue report.

This fixes serialization in the case when BackLink data has been fetched using `fetch_links=True`. In this situation, the BackLink field contains a `Document` object, which does not have the `document_class` attribute.

Thanks,

### DIFF
--- a/.github/workflows/github-actions-tests.yml
+++ b/.github/workflows/github-actions-tests.yml
@@ -19,7 +19,7 @@ jobs:
           cache: pip
           cache-dependency-path: pyproject.toml
       - name: Start MongoDB
-        uses: supercharge/mongodb-github-action@1.11.0
+        uses: supercharge/mongodb-github-action@1.12.1
         with:
           mongodb-version: ${{ matrix.mongodb-version }}
           mongodb-replica-set: test-rs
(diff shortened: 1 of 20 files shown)


CAPITAINMARVEL:
> This is my proposed solution for #1189. Please find the minimal code example in the issue report.
> 
> This fixes serialization in the case when BackLink data has been fetched using `fetch_links=True`. In this situation, the BackLink field contains a `Document` object, which does not have the `document_class` attribute.
> 
> Thanks,

maybe a getattr(self, "document_class", self.__class__) instead ?

With attribute:
try/except: 0.0754515000153333
getattr:    0.08347969996975735

Without attribute:
try/except: 0.4134541000239551
getattr:    0.08493790001375601

this is what i got with a testing script (using timeit)

roman-right:
The fix looks correct - using self.__class__ instead of the hardcoded Document class when serializing BackLinks with fetched data makes sense. However, the mergeability is currently marked as unknown, likely needs a rebase. Could you rebase against main?

CAPITAINMARVEL (approved):

PR #1209, opened 2025-07-23
fix: incomplete type hint for Regex pattern

xcrong:
Beanie's `Regex` Class can use `re.Pattern[str]` as `pattern`. In the background, `bson.regex.Regex` 's `.from_native` method support it. So I add it to type hint. 

In exist tests, we also can see that: https://github.com/BeanieODM/beanie/blob/53123ce17cb09613b0bb01a5e8ba3ab24b3e872e/tests/odm/test_encoder.py#L49-L55

image

### DIFF
--- a/beanie/odm/operators/find/evaluation.py
+++ b/beanie/odm/operators/find/evaluation.py
@@ -100,10 +100,8 @@ class RegEx(BaseFindEvaluationOperator):
     def __init__(
         self,
         field,
-        pattern: str,
-        options: Optional[
-            Union[str, re.Pattern[bytes], re.Pattern[str]]
-        ] = None,
+        pattern: Union[str, re.Pattern[bytes], re.Pattern[str]],
+        options: Optional[str] = None,
     ):
         self.field = field
         self.pattern = pattern
@@ -180,9 +178,7 @@ class Text(BaseFindEvaluationOperator):
         if self.language:
             expression["$text"]["$language"] = self.language
         if self.diacritic_sensitive is not None:
-            expression["$text"]["$diacriticSensitive"] = (
-                self.diacritic_sensitive
-            )
+            expression["$text"]["$diacriticSensitive"] = self.diacritic_sensitive
         return expression
 
 
--- a/beanie/odm/operators/find/evaluation.py
+++ b/beanie/odm/operators/find/evaluation.py
@@ -1,6 +1,6 @@
 import re
 from abc import ABC
-from typing import Optional, Union
+from typing import AnyStr, Optional, Union
 
 from beanie.odm.operators.find import BaseFindOperator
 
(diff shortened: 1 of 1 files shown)


beanie/odm/operators/find/evaluation.py
```diff
@@ -99,7 +100,7 @@ class RegEx(BaseFindEvaluationOperator):
     def __init__(
         self,
         field,
-        pattern: str,
+        pattern: Union[str, re.Pattern[str]],
```
CAPITAINMARVEL:
```suggestion
        pattern: Union[str, re.Pattern]],
```

beanie/odm/operators/find/evaluation.py
```diff
@@ -99,7 +100,7 @@ class RegEx(BaseFindEvaluationOperator):
     def __init__(
         self,
         field,
-        pattern: str,
+        pattern: Union[str, re.Pattern]],
```
staticxterm:
```suggestion
        pattern: Union[str, re.Pattern[AnyStr]],
```
And import `AnyStr` from the "typing" module.

I suggest to also modify the test "test_regex" in tests/odm/operators/find/test_evaluation.py to showcase this new usage:
```python
    pattern = re.compile("^smth")
    q = RegEx(Sample.integer, pattern)
    assert q == {"integer": {"$regex": pattern}}
```

staticxterm (changes requested):

beanie/odm/operators/find/evaluation.py
```diff
@@ -99,7 +100,7 @@ class RegEx(BaseFindEvaluationOperator):
     def __init__(
         self,
         field,
-        pattern: str,
+        pattern: Union[str, re.Pattern[AnyStr]],
```
CAPITAINMARVEL: @xcrong this time should be fine i accidentally forget a ]

CAPITAINMARVEL:
> hey, I can't understand the change, should we change it to:
> 
> ```python
>         ...
>         pattern: Union[AnyStr, re.Pattern[AnyStr]],
>         options: Optional[str] = None,
>         ...
> ```

No because pymongo will raise an error ($regex has to be a string)
pymongo can only accept a string or a re.Pattern

beanie/odm/operators/find/evaluation.py
```diff
@@ -100,7 +101,7 @@ def __init__(
         self,
         field,
         pattern: str,
-        options: Optional[str] = None,
+        options: Optional[Union[str, re.Pattern[AnyStr]]] = None,
```
CAPITAINMARVEL:
```suggestion
        options: Optional[
            Union[str, re.Pattern[bytes], re.Pattern[str]]
        ] = None,
```

beanie/odm/operators/find/evaluation.py
```diff
@@ -1,5 +1,6 @@
+import re
 from abc import ABC
-from typing import Optional
+from typing import AnyStr, Optional, Union
```
CAPITAINMARVEL:
```suggestion
from typing import Optional, Union
```

staticxterm (approved):
Okay, it took us a while, but we finally got there. Looks good to me now.

07pepa (approved):

Issue #1215, opened 2025-08-14
Add docs to Context7


staticxterm:
### Discussed in https://github.com/BeanieODM/beanie/discussions/1214

<div type='discussions-op-text'>

<sup>Originally posted by **ruze00** August 14, 2025</sup>
Adding them to context7 would allow LLM coding assistants to code Beanie much more effectively.</div>

PR #1242, opened 2025-10-20
Added context7.json #1215

gugu-py:
allow LLM coding assistants to code Beanie much more effectively #1215 

Not so sure about the `rules` field


roman-right:
I appreciate the intent, but I do not think a context7.json file belongs in the repository. LLM tooling configs are better managed externally. Closing this one.

PR #1257, opened 2025-12-14
feat: Support Python 3.14

tyzhnenko:
## Summary
 
Add support for Python 3.14 version to the test matrix and dependency bounds.

### Added
|  Dependency | Added  | Reason |
|---|---|---|
| Python 3.14  | Yes  | Released October 7, 2025 |

### Excluded
| Dependencies | Excluded | Reason
|---|---|---|
| Python 3.14 + Pydantic 2.11 | Yes | Pydantic 2.11 doesn't support Python 3.14
 
### New matrix
| Python | MongoDB | Pydantic |
|---|---|---|
| 3.10, 3.11, 3.12, 3.13, 3.14 | 5.0, 7.0, 8.0 | 2.11 (excluded for Python 3.14), 2.12 |
 
 CI jobs per PR: **27** (was 24).

### Changes
 
* `.github/workflows/github-actions-tests.yml`: updated matrix, added Python 3.14
* `pyproject.toml`: `requires-python = ">=3.10,<3.15"`, added Python 3.14 classifier
* `beanie/migrations/controllers/free_fall.py` and `beanie/migrations/controllers/iterative.py`: add support for new `inspect.signature` with [PEP 649](https://peps.python.org/pep-0649/) & [PEP 749](https://peps.python.org/pep-0749/): [Deferred evaluation of annotations](https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-deferred-annotations)
 
 
### Notes
 
* This mainly only changes the CI matrix and metadata. Added support of _Deferred evaluation of annotations_ to make sure new annotations do not break migrations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "beanie"
 version = "2.1.0"
 description = "Asynchronous Python ODM for MongoDB"
 readme = "README.md"
-requires-python = ">=3.10,<3.15"
+requires-python = ">=3.10"
 license = { file="LICENSE" }
 authors = [
     {name = "Roman Right", email = "roman-right@protonmail.com"}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "beanie"
 version = "2.1.0"
 description = "Asynchronous Python ODM for MongoDB"
 readme = "README.md"
-requires-python = ">=3.10,<3.14"
+requires-python = ">=3.10,<3.15"
 license = { file="LICENSE" }
 authors = [
     {name = "Roman Right", email = "roman-right@protonmail.com"}
@@ -28,6 +28,7 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
 ]
 dependencies = [
     "pydantic>=2.4,<3.0",
--- a/.github/workflows/github-actions-tests.yml
+++ b/.github/workflows/github-actions-tests.yml
@@ -7,11 +7,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.10", "3.11", "3.12", "3.13" ]
+        python-version: [ "3.10", "3.11", "3.12", "3.13", "3.14" ]
         # 5.0: kept for AWS DocumentDB / Azure DocumentDB compatibility
         # 7.0, 8.0: current supported MongoDB releases
         mongodb-version: [ "5.0", "7.0", "8.0" ]
         pydantic-version: [ "2.11", "2.12" ]
+        exclude:
+          - python-version: "3.14"
+            pydantic-version: "2.11"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/beanie/migrations/controllers/iterative.py
+++ b/beanie/migrations/controllers/iterative.py
@@ -1,4 +1,5 @@
 import asyncio
+import typing
 from collections.abc import Callable
 from inspect import isclass, signature
 from typing import Any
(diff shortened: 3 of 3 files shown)


staticxterm:
Hi @tyzhnenko,
thank you for this PR but there are various other issues that are not visible so to say.
One issue is that we still support Pydantic v1 and Pydantic v1 does not support Python 3.14: https://docs.pydantic.dev/1.10/

tyzhnenko:
@staticxterm I've tried to change test settings to install 1.10.25 which got support of Python 3.14 recently. But FastAPI doesn't work in this scenario [1].

1 – https://github.com/BeanieODM/beanie/actions/runs/20381641926/job/58573292623?pr=1257

tyzhnenko:
Probably should be merged after #1170

roman-right:
The direction is right - we should add Python 3.14 support and drop 3.9 at some point. This was marked stale. If you are still interested, please rebase against current main and I will review.

staticxterm:
Anyway, I am against merging this change, simply on the basis that it is too simple, and there are other hidden issues not popping up with our tests scope. That's why I've explicitly pinned the upper range of supported Python versions to below 3.14.

This PR could instead focus on adding true support for Python 3.14, whatever is needed to fully support it apart from simply stating so in pyproject.toml and in the test matrix and so on.
For example, what I'm aware of is that Beanie in some places does not handle annotation unpacking/introspection corretly (basically what is described as "deferred annotations" [here](https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-deferred-annotations), so that would need to be aligned with how for example FastAPI fixed it (one example I came across [here](https://github.com/fastapi/fastapi/pull/14789/changes)).

tyzhnenko:
@staticxterm it seems FastAPI with py314 doesn't really work with Pydantic v1 at all. Tests are failed due to this line [here](https://github.com/fastapi/fastapi/blob/0.124.4/fastapi/_compat/may_v1.py#L59)

I'll prepare a patch for `signature` usage and it could be merged after #1287

staticxterm:
> @staticxterm it seems FastAPI with py314 doesn't really work with Pydantic v1 at all. Tests are failed due to this line [here](https://github.com/fastapi/fastapi/blob/0.124.4/fastapi/_compat/may_v1.py#L59)
> 
> I'll prepare a patch for `signature` usage and it could be merged after #1287

Interesting. Thank you for looking into this. It seems that we have to drop support for Pydantic v1 then since even FastAPI is not compatible with it anymore. 

@roman-right If it's okay with you, I'd like to work on a PR regarding that but as tyzhnenko has stated here, first step is to merge #1287.

tyzhnenko:
> Interesting. Thank you for looking into this. It seems that we have to drop support for Pydantic v1 then since even FastAPI is not compatible with it anymore.
> 
> @roman-right If it's okay with you, I'd like to work on a PR regarding that but as tyzhnenko has stated here, first step is to merge #1287.

Yep, FastAPI changed dependency in 0.126 (see [here](https://github.com/fastapi/fastapi/blob/0.126.0/pyproject.toml)) and completely dropped Pydantic v1 in 0.128, see release notes [here](https://fastapi.tiangolo.com/release-notes/#01280)

tyzhnenko:
@roman-right @staticxterm take a look please

roman-right:
Reworked the PR. Changes:

- Removed the duplicated signature wrapper from both migration files
- No changes to free_fall.py (it only checks parameter names, not annotations)
- iterative.py now uses typing.get_type_hints() to resolve annotations, which handles deferred annotations on all Python versions without version-specific code
- No unrelated formatting changes to pyproject.toml
- Rebased on current main

pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "beanie"
 version = "2.1.0"
 description = "Asynchronous Python ODM for MongoDB"
 readme = "README.md"
-requires-python = ">=3.10,<3.14"
+requires-python = ">=3.10,<3.15"
```
staticxterm: The correct version range should be unconstrained:
```suggestion
requires-python = ">=3.10"
```

staticxterm (commented):
One small comment regarding the current changes for now.

@roman-right there are no actual tests that would verify that Python 3.14 "deferred annotations" support works correctly or not - they should be added so that we are confident that these changes are enough.

I believe that more changes may be needed, e.g. in the Link/BackLink classes (`document_class` handling) and probably also in the `DocsRegistry.evaluate_fr()`.

AL-Troy:
@roman-right Just wanted to check in on this PR to see if it's still on your radar and if there's anything blocking it at the moment. Please let me know if there's anything I can do to help move the ticket forward. Thanks!

PR #1304, opened 2026-03-16
added beanie sort support for find_one

CAPITAINMARVEL:


Riverfount (dismissed):
LGTM

beanie/odm/interfaces/find.py
```diff
@@ -53,6 +53,7 @@ def find_one(  # type: ignore
         cls: type[FindType],
         *args: Mapping[Any, Any] | bool,
         projection_model: None = None,
+        sort: Any | list[tuple[Any, SortDirection]] | None = None,
```
staticxterm: This `Any | list[...] | None = None` typing will simply result in `Any`, since it "trumps" all the rest. The other types will never be checked.
Unfortunately, since we can't enforce stricter typing due to the `ExpressionField` masking it for type checkers, typing it as `Any` is the best we can do.
```suggestion
        sort: Any = None,
```
Same comment for every other occurence in this file.

beanie/odm/queries/find.py
```diff
@@ -189,7 +216,7 @@ def find_many(
         projection_model: None = None,
         skip: int | None = None,
         limit: int | None = None,
-        sort: None | str | list[tuple[str, SortDirection]] = None,
+        sort: None | Any | list[tuple[Any, SortDirection]] = None,
```
staticxterm:
```suggestion
        sort: Any = None,
```
`Any` trumps all other types so the other types will never get checked.
Same comment for every occurrence in this file.

beanie/odm/documents.py
```diff
@@ -1342,7 +1342,7 @@ def find_many_in_all(  # type: ignore
         projection_model: type["DocumentProjectionType"] | None = None,
         skip: int | None = None,
         limit: int | None = None,
-        sort: None | str | list[tuple[str, SortDirection]] = None,
+        sort: None | Any | list[tuple[Any, SortDirection]] = None,
```
staticxterm:
```suggestion
        sort: Any = None,
```
Same comment for all other occurrences in this file.

beanie/odm/queries/find.py
```diff
@@ -52,6 +52,33 @@
...
+        elif isinstance(arg, str):
+            if arg.startswith("+"):
+                sort_list.append((arg[1:], SortDirection.ASCENDING))
+            elif arg.startswith("-"):
+                sort_list.append((arg[1:], SortDirection.DESCENDING))
+            else:
+                sort_list.append((arg, SortDirection.ASCENDING))
+        else:
+            raise TypeError("Wrong argument type")
+    return sort_list
```
staticxterm, line 79: Propose to not expose this method as part of the public API but to make it "private"/internal by prefixing it with an underscore.
I played around with this a bit - my proposal for this method using the new structural matching in Python 3.10:

```suggestion
def _build_sort_list(
    *args: Any,
) -> list[tuple[str, SortDirection]]:
    sort_list: list[tuple[str, SortDirection]] = []
    if not args and all(a is None for a in args):
        raise TypeError(
            "sort() requires at least one argument. "
            "Expected: "
            "str (e.g. '-field'), "
            "tuple[str, SortDirection] (e.g. ('field', SortDirection.DESCENDING)), "
            "tuple[str, int] (e.g. ('field', -1)), "
            "or a list of those."
        )

    for i, arg in enumerate(args):
        match arg:
            case list():
                sort_list.extend(_build_sort_list(*arg))

            case (str() as field, SortDirection() as direction):
                # Matches both:
                #   tuple[ExpressionField, SortDirection], e.g. -Sample.price
                #   tuple[str, SortDirection],             e.g. ("price", SortDirection.DESCENDING)
                sort_list.append((field, direction))

            case (str() as field, int() as direction):
                # Matches both:
                #   tuple[ExpressionField, int], e.g. -Sample.price
                #   tuple[str, int],             e.g. ("price", -1) or ("price", pymongo.DESCENDING)
                #
                # Important: this case must stay after the SortDirection() case,
                # because SortDirection is an IntEnum, so its values also match int()
                # and would otherwise be handled here instead.
                if direction not in (-1, 1):
                    raise TypeError(
                        f"sort() argument {i} has invalid direction {direction!r} "
                        f"for field {field!r}. "
                        "Expected direction to be either 1 (ascending) or -1 (descending)."
                    )

                sort_list.append(
                    (
                        field,
                        SortDirection.DESCENDING
                        if direction == -1
                        else SortDirection.ASCENDING,
                    )
                )

            case str() as field:
                if field.startswith("+"):
                    sort_list.append((field[1:], SortDirection.ASCENDING))
                elif field.startswith("-"):
                    sort_list.append((field[1:], SortDirection.DESCENDING))
                else:
                    sort_list.append((field, SortDirection.ASCENDING))

            case _:
                raise TypeError(
                    f"sort() argument {i} has unsupported type {type(arg).__name__!r}. "
                    "Expected: "
                    "str (e.g. '-field'), "
                    "tuple[str, SortDirection] (e.g. ('field', SortDirection.DESCENDING)), "
                    "tuple[str, int] (e.g. ('field', -1)), "
                    "or a list of those. "
                    f"Got: {arg!r}"
                )

    return sort_list
```
It also adds a bit more informative message to the user what went wrong, since we now rely on this runtime type of checking as we can't satisfy it statically with type checkers now that sort argument is typed as Any.

beanie/odm/queries/find.py
```diff
@@ -364,12 +391,10 @@ def find(
...
-        *args: (
-            str
-            | tuple[str, SortDirection]
-            | list[tuple[str, SortDirection]]
-            | None
-        ),
+        *args: Any
+        | tuple[Any, SortDirection]
+        | list[tuple[Any, SortDirection]]
+        | None,
```
staticxterm, line 398: Again, type it as `Any`.
```suggestion
        *args: Any,
```

beanie/odm/queries/find.py
```diff
@@ -379,28 +404,7 @@ def sort(
...
-                    self.sort_expressions.append(
-                        (arg[1:], SortDirection.DESCENDING)
-                    )
-                else:
-                    self.sort_expressions.append(
-                        (arg, SortDirection.ASCENDING)
-                    )
-            else:
-                raise TypeError("Wrong argument type")
+        self.sort_expressions.extend(build_sort_list(*args))
```
staticxterm, line 408: Reaching this method and this part of code currently also happens through e.g. `Sample.find_many(Sample.integer > 0)`, because the `find_many()` `sort` argument defaults to None.
We should do a little bugfix here for the FindMany.find_many() method to not call this `sort()` method if `sort=None`...

beanie/odm/queries/find.py
staticxterm: _A little bugfix_
Similarly as you did for `find_one()` below, we should prevent calling of `_build_sort_list()` when `sort=None` by default (i.e., argument not provided).
```suggestion
        if sort is not None:
            self.sort(sort)
```

beanie/odm/queries/find.py
```diff
@@ -799,13 +806,17 @@ def find_one(
...
+        or a list of (key, direction) pairs specifying the sort order
+        for this query.
         :param session: Optional[AsyncClientSession] - pymongo session
         :param ignore_cache: bool
         :param **pymongo_kwargs: pymongo native parameters for find operation (if Document class contains links, this parameter must fit the respective parameter of the aggregate MongoDB function)
         :return: FindOne - query instance
         """
         self.find_expressions += args  # type: ignore # bool workaround
         self.project(projection_model)
+        self.sort_expressions = build_sort_list(sort) if sort else []
```
staticxterm, line 820: Please also add a few unit tests covering this change, e.g. here I focused on the `find_many().sort()` but you can extend to include `find_one(..., sort=...)` case(s) as well:
tests/odm/query/test_sort.py
```python
"""
Tests for the .sort() method on FindMany (and the sort= kwarg on find_many/find_one).

Covered cases
─────────────
String forms         : bare "field", ascending "+field", descending "-field"
ExpressionField      : unary +Sample.field (ASC) and -Sample.field (DESC)
List argument        : list of (ExpressionField, SortDirection) tuples
Mixed positional     : multiple positional args of different types
sort= kwarg          : passed directly to find_many() / find_all()
sort_expressions     : internal state after calling .sort()
Chaining             : multiple .sort() calls accumulate correctly
Clone isolation      : .clone() does not share sort state with the original
Invalid input        : no args, non-string / non-tuple args raise TypeError
Real query ordering  : documents returned in the expected order (async)
"""

import pymongo
import pytest

from beanie.odm.enums import SortDirection
from tests.odm.models import Sample


class TestSortExpressionsFromStringForms:
    """FindMany.sort() called with plain string arguments."""

    def test_bare_field_name_is_ascending(self):
        q = Sample.find_many(Sample.integer > 0).sort("integer")
        assert q.sort_expressions == [("integer", SortDirection.ASCENDING)]

    def test_plus_prefix_is_ascending(self):
        q = Sample.find_many(Sample.integer > 0).sort("+integer")
        assert q.sort_expressions == [("integer", SortDirection.ASCENDING)]

    def test_minus_prefix_is_descending(self):
        q = Sample.find_many(Sample.integer > 0).sort("-integer")
        assert q.sort_expressions == [("integer", SortDirection.DESCENDING)]

    def test_nested_field_bare(self):
        q = Sample.find_many(Sample.integer > 0).sort("nested.integer")
        assert q.sort_expressions == [
            ("nested.integer", SortDirection.ASCENDING)
        ]

    def test_nested_field_descending(self):
        q = Sample.find_many(Sample.integer > 0).sort("-nested.integer")
        assert q.sort_expressions == [
            ("nested.integer", SortDirection.DESCENDING)
        ]

    def test_multiple_string_args(self):
        q = Sample.find_many(Sample.integer > 0).sort("-integer", "+string")
        assert q.sort_expressions == [
            ("integer", SortDirection.DESCENDING),
            ("string", SortDirection.ASCENDING),
        ]

    def test_sort_kwarg_bare_string(self):
        q = Sample.find_many(Sample.integer > 0, sort="integer")
        assert q.sort_expressions == [("integer", SortDirection.ASCENDING)]

    def test_sort_kwarg_descending_string(self):
        q = Sample.find_many(Sample.integer > 0, sort="-integer")
        assert q.sort_expressions == [("integer", SortDirection.DESCENDING)]


class TestSortExpressionsFromExpressionFields:
    """FindMany.sort() called with ExpressionField unary operators."""

    def test_positive_unary_is_ascending(self):
        q = Sample.find_many(Sample.integer > 0).sort(+Sample.integer)
        assert q.sort_expressions == [("integer", SortDirection.ASCENDING)]

    def test_negative_unary_is_descending(self):
        q = Sample.find_many(Sample.integer > 0).sort(-Sample.integer)
        assert q.sort_expressions == [("integer", SortDirection.DESCENDING)]

    def test_positive_unary_nested_field(self):
        q = Sample.find_many(Sample.integer > 0).sort(+Sample.nested.integer)
        assert q.sort_expressions == [
            ("nested.integer", SortDirection.ASCENDING)
        ]

    def test_negative_unary_nested_field(self):
        q = Sample.find_many(Sample.integer > 0).sort(-Sample.nested.integer)
        assert q.sort_expressions == [
            ("nested.integer", SortDirection.DESCENDING)
        ]

    def test_multiple_expression_field_args(self):
        q = Sample.find_many(Sample.integer > 0).sort(
            -Sample.integer, +Sample.string
        )
        assert q.sort_expressions == [
            ("integer", SortDirection.DESCENDING),
            ("string", SortDirection.ASCENDING),
        ]

    def test_sort_kwarg_expression_field_descending(self):
        q = Sample.find_many(Sample.integer > 0, sort=-Sample.integer)
        assert q.sort_expressions == [("integer", SortDirection.DESCENDING)]

    def test_sort_kwarg_expression_field_ascending(self):
        q = Sample.find_many(Sample.integer > 0, sort=+Sample.integer)
        assert q.sort_expressions == [("integer", SortDirection.ASCENDING)]


class TestSortExpressionsFromListArgument:
    """FindMany.sort() called with a single list-of-tuples argument."""

    def test_list_of_expression_field_tuples(self):
        q = Sample.find_many(Sample.integer > 0).sort(
            [Sample.const, -Sample.integer]
        )

        assert q.sort_expressions == [
            ("const", SortDirection.ASCENDING),
            ("integer", SortDirection.DESCENDING),
        ]

    def test_list_of_string_direction_tuples(self):
        q = Sample.find_many(Sample.integer > 0).sort(
            [("integer", pymongo.DESCENDING), ("string", pymongo.ASCENDING)]
        )
        assert q.sort_expressions == [
            ("integer", SortDirection.DESCENDING),
            ("string", SortDirection.ASCENDING),
        ]

    def test_sort_kwarg_list_of_string_tuples(self):
        q = Sample.find_many(
            Sample.integer > 0,
            sort=[("integer", pymongo.DESCENDING)],
        )
        assert q.sort_expressions == [("integer", SortDirection.DESCENDING)]


class TestSortExpressionsMixedPositionalArgs:
    """FindMany.sort() with a mix of string and ExpressionField positional args."""

    def test_string_then_expression_field(self):
        q = Sample.find_many(Sample.integer > 0).sort(
            "-integer", +Sample.string
        )
        assert q.sort_expressions == [
            ("integer", SortDirection.DESCENDING),
            ("string", SortDirection.ASCENDING),
        ]

    def test_expression_field_then_string(self):
        q = Sample.find_many(Sample.integer > 0).sort(
            -Sample.integer, "+string"
        )
        assert q.sort_expressions == [
            ("integer", SortDirection.DESCENDING),
            ("string", SortDirection.ASCENDING),
        ]


class TestSortChaining:
    """Multiple FindMany.sort() calls on the same query object accumulate."""

    def test_two_sort_calls_accumulate(self):
        q = (
            Sample.find_many(Sample.integer > 0)
            .sort("-integer")
            .sort("+string")
        )
        assert q.sort_expressions == [
            ("integer", SortDirection.DESCENDING),
            ("string", SortDirection.ASCENDING),
        ]

    def test_three_sort_calls_accumulate(self):
        q = (
            Sample.find_many(Sample.integer > 0)
            .sort("-integer")
            .sort("+string")
            .sort("float_num")
        )
        assert q.sort_expressions == [
            ("integer", SortDirection.DESCENDING),
            ("string", SortDirection.ASCENDING),
            ("float_num", SortDirection.ASCENDING),
        ]

    def test_chained_sort_does_not_affect_earlier_reference(self):
        base = Sample.find_many(Sample.integer > 0).sort("-integer")
        # Obtain a reference to the current list length before chaining
        expressions_before = list(base.sort_expressions)
        base.sort("+string")
        # The base query object itself gets mutated (sort returns self);
        # assert the first expression is still correct
        assert expressions_before[0] == ("integer", SortDirection.DESCENDING)


class TestSortCloneIsolation:
    """Cloning a query produces an independent copy of sort_expressions."""

    def test_clone_inherits_sort_expressions(self):
        q = Sample.find_many(Sample.integer > 0).sort("-integer")
        cloned = q.clone()
        assert cloned.sort_expressions == [
            ("integer", SortDirection.DESCENDING)
        ]

    def test_clone_mutation_does_not_affect_original(self):
        q = Sample.find_many(Sample.integer > 0).sort("-integer").limit(100)
        cloned = q.clone()
        cloned.sort("+string")

        # original unchanged
        assert q.sort_expressions == [("integer", SortDirection.DESCENDING)]
        # clone has both
        assert cloned.sort_expressions == [
            ("integer", SortDirection.DESCENDING),
            ("string", SortDirection.ASCENDING),
        ]

    def test_original_mutation_does_not_affect_clone(self):
        q = Sample.find_many(Sample.integer > 0).sort("-integer").limit(100)
        cloned = q.clone()
        q.sort("+float_num")

        assert cloned.sort_expressions == [
            ("integer", SortDirection.DESCENDING)
        ]
        assert q.sort_expressions == [
            ("integer", SortDirection.DESCENDING),
            ("float_num", SortDirection.ASCENDING),
        ]


class TestSortInvalidInput:
    """Invalid sort arguments raise TypeError."""

    def test_no_args_raises_type_error(self):
        with pytest.raises(
            TypeError, match="sort\\(\\) requires at least one argument"
        ):
            Sample.find_many(Sample.integer > 0).sort()

    def test_integer_arg_raises_type_error(self):
        with pytest.raises(TypeError):
            Sample.find_many(Sample.integer > 0, sort=1)

    def test_integer_passed_to_sort_method_raises_type_error(self):
        with pytest.raises(TypeError):
            Sample.find_many(Sample.integer > 0).sort(1)


class TestSortRealOrdering:
    """Verify actual document ordering returned from MongoDB."""

    async def test_string_descending_orders_results(self, preset_documents):
        result = await Sample.find_many(
            Sample.integer > 1, sort="-integer"
        ).to_list()
        for prev, curr in zip(result, result[1:]):
            assert prev.integer >= curr.integer

    async def test_string_ascending_orders_results(self, preset_documents):
        result = await Sample.find_many(
            Sample.integer > 1, sort="+integer"
        ).to_list()
        for prev, curr in zip(result, result[1:]):
            assert prev.integer <= curr.integer

    async def test_bare_string_ascending_orders_results(
        self, preset_documents
    ):
        result = await Sample.find_many(
            Sample.integer > 1, sort="integer"
        ).to_list()
        for prev, curr in zip(result, result[1:]):
            assert prev.integer <= curr.integer

    async def test_expression_field_descending_orders_results(
        self, preset_documents
    ):
        result = await Sample.find_many(
            Sample.integer > 1, sort=-Sample.integer
        ).to_list()
        for prev, curr in zip(result, result[1:]):
            assert prev.integer >= curr.integer

    async def test_expression_field_ascending_orders_results(
        self, preset_documents
    ):
        result = await Sample.find_many(
            Sample.integer > 1, sort=+Sample.integer
        ).to_list()
        for prev, curr in zip(result, result[1:]):
            assert prev.integer <= curr.integer

    async def test_list_of_expression_field_tuples_orders_results(
        self, preset_documents
    ):
        result = (
            await Sample.find_many(Sample.integer > 1)
            .sort([Sample.const, -Sample.integer])
            .to_list()
        )
        # const is uniform ("TEST") so secondary sort drives ordering
        for prev, curr in zip(result, result[1:]):
            assert prev.integer >= curr.integer

    async def test_multiple_positional_expression_fields_order_results(
        self, preset_documents
    ):
        result = (
            await Sample.find_many(Sample.integer > 0)
            .sort(-Sample.integer, +Sample.increment)
            .to_list()
        )
        # Primary: integer DESC
        for prev, curr in zip(result, result[1:]):
            assert prev.integer >= curr.integer

    async def test_sort_chained_with_limit(self, preset_documents):
        result = (
            await Sample.find_many(Sample.increment > 2)
            .find_many(Sample.nested.optional == None)
            .sort(Sample.increment)
            .limit(2)
            .to_list()
        )
        assert len(result) == 2
        assert result[0].increment <= result[1].increment

    async def test_sort_via_find_all_kwarg(self, preset_documents):
        import pymongo

        result = await Sample.find_all(
            sort=[
                ("string", pymongo.ASCENDING),
                ("integer", pymongo.DESCENDING),
            ]
        ).to_list()
        # Primary key ascending
        for prev, curr in zip(result, result[1:]):
            assert prev.string <= curr.string

    async def test_first_or_none_with_sort_returns_max(self, preset_documents):
        doc = (
            await Sample.find(Sample.increment > 1)
            .sort(-Sample.increment)
            .first_or_none()
        )
        assert doc is not None
        assert doc.increment == 9
```

staticxterm (changes requested):

roman-right (changes requested):
The feature itself is useful -- `find_one(sort=...)` is a common MongoDB pattern and it was missing from Beanie. The structural approach (extracting sort parsing into a shared function) is the right call.

Please address all of @staticxterm's review comments. They are correct across the board.

A few additional things I noticed:

1. `FindOne` has no `sort_expressions` attribute in `__init__`. It is only initialized in `FindMany.__init__()`. If `FindOne.__await__` runs on an instance that never went through `find_one()`, accessing `self.sort_expressions` will raise `AttributeError`. Initialize it in `FindQuery.__init__()` or in `FindOne.__init__()`.

2. The cache key in `FindOne.__await__()` includes `self.sort_expressions` as a list. Lists are mutable and unhashable. Convert to tuple before passing to `LRUCache.create_key()`: `tuple(self.sort_expressions)`.

3. The docstring says `Union[None, str, list[tuple[str, SortDirection]]]` but the actual annotation is `Any | list[tuple[Any, SortDirection]]`. These should match. If you simplify to `Any` per @staticxterm's suggestion, describe the accepted runtime forms in the docstring instead of repeating a type annotation.

4. Tests are required. @staticxterm provided a solid test suite in comment #8 -- use it as a starting point.

beanie/odm/documents.py
```diff
@@ -1380,7 +1380,7 @@ def find_many(  # type: ignore
...
```
07pepa, line 1383: sort: Any = None, shuld be here as well

07pepa (changes requested):

Issue #163, opened 2021-12-11
Suggestion for dynamic projections


gethvi:
Hi,

I found a use case where I need to have projections made dynamically, basically just by specifying an arbitrary subset of fields that I want. My solution was to dynamically create new Model and pass it to the `project()` function. I was wondering if you would accept something like this as a PR? Would it be useful for Beanie? Or would you suggest another way of doing this? Thanks!

Right now it I am using it like this:

```python
Product.find_many().project(create_submodel(Product, fields=["name", "description"]))
```

If made as part of Beanie, could it be like this?
```python
Product.find_many().project(fields=["name", "description"])
```

PoC source with an example:

```python
from typing import List, Union, Dict, Type, Optional
from pydantic import create_model, BaseModel


def create_submodel(
    model_type: Type[BaseModel], fields: List[Union[str, Dict]]
) -> Type[BaseModel]:
    """
    Creates a new model based on specified model_type with only a subset of it's fields.

    :param model_type:
    :param fields:
    :return:
    """
    model_fields = model_type.__fields__
    model_field_names = set(model_fields.keys())
    submodel_fields = {}
    for field in fields:

        if isinstance(field, str):

            if field not in model_field_names:
                raise ValueError(
                    f"field '{field}' not defined in model {model_type.__name__}"
                )

            submodel_fields[field] = (
                model_fields[field].type_,
                model_fields[field].field_info,
            )

        elif isinstance(field, dict):
            field.values()
            field_name = list(field.keys())[0]

            if field_name not in model_field_names:
                raise ValueError(
                    f"field '{field_name}' not defined in model {model_type.__name__}"
                )

            submodel = create_submodel(
                model_fields[field_name].type_, field[field_name]
            )
            submodel_fields[field_name] = (submodel, model_fields[field_name].field_info)

    return create_model(f"{model_type.__name__}SubModel", **submodel_fields)


if __name__ == "__main__":

    class Category(BaseModel):
        name: str
        description: str

    class Product(BaseModel):
        name: str
        description: Optional[str] = None
        price: float
        category: Category

    ProductSubModel = create_submodel(
        Product, fields=["name", "price", {"category": ["name"]}]
    )

    product_with_less_info = ProductSubModel.parse_obj(
        {"name": "Banana", "price": 3.14, "category": {"name": "Fruit"}}
    )

    print(product_with_less_info.json(indent=2))

# RESULT:
# {
#   "name": "Banana",
#   "price": 3.14,
#   "category": {
[...]

PR #1322, opened 2026-03-29
feat: add .exclude() method for exclusion projection on find queries

smahn9123:
Beanie's `projection_model` only supports inclusion — you define a model listing fields you want. But sometimes you just want to skip a couple of fields (a blob, sensitive data, etc.) without spelling out everything else. MongoDB supports this natively with `{field: 0}`, so this PR exposes it:

```python
products = await Product.find().exclude("internal_notes", "api_key").to_list()
```

Also works with `find_one`, `fetch_links`, and ExpressionField references:

```python
product = await Product.find_one(Product.id == pid).exclude("internal_notes")
products = await Product.find(fetch_links=True).exclude(Product.internal_notes).to_list()
```

Multiple calls accumulate (`.exclude("a").exclude("b")` excludes both), and combining with `.project()` raises `ValueError` since MongoDB doesn't allow mixing inclusion/exclusion.

Related: #163

---

### Implementation notes

The tricky part was parsing. MongoDB returns documents without excluded fields, but `model_validate` fails on missing required fields. The fix: dynamically create a subclass where excluded fields become `Optional[T] = None`, cached via `lru_cache`. The subclass still passes `isinstance` checks.

For `fetch_links=True` (aggregation pipeline), exclusion uses `$unset` instead of `$project` — `$project` with `{field: 0}` is unreliable for non-`_id` fields depending on MongoDB version.

Field names (Python names, MongoDB aliases, ExpressionField references) are all normalised to Python field names at `.exclude()` call time, then converted to MongoDB names only when building the query. This keeps resolution in one place (`_to_python_name`) rather than scattered across query and parsing paths.

> Excluded fields become `None` on returned documents. Saving them back would overwrite real values — intended for read-only use.

### Changes

- `beanie/odm/utils/projection.py` — `get_exclusion_model()`: cached dynamic model creation
- `beanie/odm/queries/find.py` — `.exclude()` on FindQuery/FindMany/FindOne, field name normalisation, `$unset` for aggregation, cache keys
- `docs/tutorial/find.md` — documentation
- `tests/odm/query/test_find.py` — 8 tests (all paths, 577 existing tests pass)

### DIFF
--- a/beanie/odm/queries/find.py
+++ b/beanie/odm/queries/find.py
@@ -39,7 +39,8 @@ from beanie.odm.utils.dump import get_dict
 from beanie.odm.utils.encoder import Encoder
 from beanie.odm.utils.find import construct_lookup_queries, split_text_query
 from beanie.odm.utils.parsing import parse_obj
-from beanie.odm.utils.projection import get_projection
+from beanie.odm.utils.projection import get_exclusion_model, get_projection
+from beanie.odm.utils.pydantic import get_model_fields
 from beanie.odm.utils.relations import resolve_query_paths
 
 if TYPE_CHECKING:
@@ -80,6 +81,7 @@ class FindQuery(
         self.lazy_parse = False
         self.nesting_depth: int | None = None
         self.nesting_depths_per_field: dict[str, int] | None = None
+        self._exclude_fields: list[str] = []
 
     def prepare_find_expressions(self):
         if self.document_model.get_link_fields() is not None:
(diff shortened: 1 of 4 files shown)


beanie/odm/queries/find.py
```diff
@@ -124,18 +126,109 @@ def delete(
...
+    # Data flow:
+    #   .exclude("id", Sample.float_num, "_id")
+    #       ↓  _to_python_name() normalises each input
+    #   _exclude_fields = ["id", "float_num"]   (canonical Python names)
+    #       ↓                          ↓
+    #   _excluded_db_fields()      get_exclusion_model()
+    #   → ["_id", "float_num"]     → model with Optional fields
+    #   (for MongoDB query)        (for Pydantic parsing)
+
+    def exclude(self, *fields: str) -> "FindQuery[FindQueryResultType]":
```
staticxterm: `*fields: str` - this type is not correct.
"*fields" is a tuple of non-keyword arguments.

smahn9123:
will work on this later.

Issue #836, opened 2024-01-23
[BUG] mypy strict no-untyped-call


aaronted009:
**Describe the bug**
When using Beanie with mypy strict typing, some typing issues make it impossible to pass the linter validation.

**To Reproduce**
Here is an example of code:
```python
import asyncio
from typing import Optional

from motor.motor_asyncio import AsyncIOMotorClient
from pydantic import BaseModel

from beanie import Document, init_beanie, operators


class Category(BaseModel):
    name: str
    description: str


class Product(Document):
    name: str
    description: Optional[str] = None
    price: float
    category: Category


async def example() -> None:
    client = AsyncIOMotorClient("mongodb://user:pass@host:27017")
    await init_beanie(database=client["beanie_test"], document_models=[Product])

    chocolate = Category(
        name="Chocolate",
        description="A preparation of roasted and ground cacao seeds.",
    )
    tonybar = Product(name="Tony's", price=5.95, category=chocolate)
    await tonybar.insert()

    product = await Product.find_one(operators.LTE(Product.price, 10))


if __name__ == "__main__":
    asyncio.run(example())
```
Here is the output.
```
$ > mypy --stritct _local/test_mypy_1.py
test_mypy_1.py:35: error: Call to untyped function "LTE" in typed context  [no-untyped-call]
```
**Expected behavior**
There should be no error

I believe this can be fixed by specifying the NoneType for the functions missing it. 

If no objection, I can submit a Pull Request for the changes.

roman-right:
Thank you for the catch! I'll add this case to the typing tests suite

aaronted009:
Can I submit a PR for this

joaoheusi:
I would also like to see this implemented. It bugs me to have to use `# type: ignore` every time i use a operator

PR #925, opened 2024-04-30
Add missing type signature to `BaseFindComparisonOperator` constructor

aaronted009:
Fix #836 , #672

PR #1329, opened 2026-05-31
fix: type-annotate update general operators for mypy --strict

### LINKED ISSUE #946: [BUG]
**Describe the bug**
  This is the same basic bug as #836, except it applies to `BaseUpdateGeneralOperator` instead of `BaseFindComparisonOperator`.

**To Reproduce**
Running `mypy` in strict mode produces `error: Call to untyped function "Set" in typed context  [no-untyped-call]` for the call to `Set` on line 10 of the following Python code:
```python
from beanie import Document


class Sample(Document):
    one: int


def main() -> None:
    Set({Sample.one: 2})
```

**Expected behavior**
There should be no `mypy` errors.

**Additional context**
Pull request #925 will fix issue #836, but it will not fix this one; fixing this issue would require also type-hinting the return value of `BaseUpdateGeneralOperator.__init__()`. It might be worth diving into the code to see if there are other similar instances, although I don't know of any right now.

Thanks!

estelledc:
Closes #946.

## Problem

\`BaseUpdateGeneralOperator.__init__\` and \`SetRevisionId.__init__\` were untyped, so calling \`Set({Sample.one: 2})\` in a \`mypy --strict\` context fires:

\`\`\`
error: Call to untyped function \"Set\" in typed context  [no-untyped-call]
\`\`\`

This is the sibling issue of #836 (which was fixed by #925 for the find comparison operators) but on the update operator side.

## Fix

Annotate the \`__init__\` parameters and the \`query\` property's return type. I used \`Mapping[Any, Any]\` for the \`expression\` argument rather than the more obvious \`Mapping[str, Any]\` so callers using \`ExpressionField\` keys (which subclass \`str\` but are typed as the original field type, e.g. \`int\` in the snippet from the issue) still type-check without spurious \`dict-item\` errors.

## Verification

Reproduced the original error and the fix locally with \`mypy --strict\` against the snippet from the issue body:

\`\`\`bash
# without the fix
$ mypy --strict test_typing_set.py
test_typing_set.py:10: error: Call to untyped function \"Set\" in typed context  [no-untyped-call]
Found 1 error in 1 file (checked 1 source file)

# with the fix
$ mypy --strict test_typing_set.py
Success: no issues found in 1 source file
\`\`\`

7 lines added, 4 lines changed.

### DIFF
--- a/beanie/odm/operators/update/general.py
+++ b/beanie/odm/operators/update/general.py
@@ -1,14 +1,17 @@
+from collections.abc import Mapping
+from typing import Any
+
 from beanie.odm.operators.update import BaseUpdateOperator
 
 
 class BaseUpdateGeneralOperator(BaseUpdateOperator):
     operator = ""
 
-    def __init__(self, expression):
+    def __init__(self, expression: Mapping[Any, Any]) -> None:
         self.expression = expression
 
     @property
-    def query(self):
+    def query(self) -> Mapping[str, Any]:
         return {self.operator: self.expression}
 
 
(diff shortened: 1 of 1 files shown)


roman-right (approved):
LGTM. Clean, minimal type annotation fix. Good call using Mapping[Any, Any] for xpression to accommodate ExpressionField typed keys. Fixes #946.

Issue #1050, opened 2024-10-10
Beanie is fully type annotated


staticxterm:
Issue to track progress on adding or fixing type annotations in Beanie project.
The goal is to have Beanie fully (and correctly!) type annotated (where possible).

Issues at hand:
- [ ] Kwargs and args not type annotated
- [ ] Add full type annotations to Beanie migrations code:
  - [ ] beanie/executors/migrate.py
  - [ ] beanie/migrations/controllers/free_fall.py
  - [ ] beanie/migrations/controllers/iterative.py
- [ ] Fix Generic type annotations:
  - [ ] beanie/odm/union_doc.py (`cls._document_models` and `doc_model` Generic typing)
  - [ ] beanie/odm/fields.py (issues in `Link` class)

_Work in progress, more issues/categories to be added._

ttamg:
I keep running into the id field type checking issue that is covered in this old issue #60 

We use `UUID` for the `id` field in Mongo / Beanie rather than the `PydanticObjectID`

Certainly in the `Document` definition, allowing `UUID` as an alternative type would I believe allow the type checker to do its work.

```
class Document 
...

    id: Optional[PydanticObjectId | UUID] = Field(
        default=None, description="MongoDB document ObjectID"
    )
```

Any progress on plans to allow `UUID` in the `id` field for typechecking?  Thanks

filipchristiansen:
**`init_beanie(database=...)` type error with Motor**

In [`beanie/odm/utils/init.py`](https://github.com/BeanieODM/beanie/blob/5569cdbd1a26f0af480e6d8e8cd533802b0572f6/beanie/odm/utils/init.py#L751-L752), `init_beanie`’s `database` parameter is annotated as `Optional[AsyncDatabase]` (from `pymongo.asynchronous.database`). In normal Motor usage you pass the result of `AsyncIOMotorClient(...).get_default_database()`, which Motor’s stubs type as `AsyncIOMotorDatabase[_DocumentType]` ([`motor_asyncio.pyi` L47](https://github.com/mongodb/motor/blob/4fe13c6f254158abae461b1539c08358f2562b10/motor/motor_asyncio.pyi#L47)); in practice the checker reports it as `AsyncIOMotorDatabase[Any]`. Type checker (`mypy`) then reports:

```
error: Argument "database" to "init_beanie" has incompatible type "AsyncIOMotorDatabase[Any]"; expected "AsyncDatabase[Any] | None"  [arg-type]
```

[...]

PR #1330, opened 2026-05-31
fix(typing): annotate leftover params on find comparison operators

### LINKED ISSUE #946: [BUG]

estelledc:
Refs #1050.

The tracking issue lists \`beanie/odm/operators/find/comparison.py\` as a remaining typing leftover from #925, which only added \`-> None\` to \`BaseFindComparisonOperator.__init__\`. The \`field\` and \`other\` parameters and the \`query\` property return type were still bare.

This PR annotates explicitly:

- \`field: Any\`, \`other: Any\` on the constructor.
- \`query\` properties return \`Mapping[str, Any]\` on the base class and on the \`Eq\` override that has its own implementation.

Mirrors the shape we just used for the update-side operators in #1329 (which closed #946, the sibling issue). No runtime change; \`mypy --strict\` treats the explicit annotations the same as the previous implicit-Any, but they're now visible to readers and to stricter checkers (pyright in strict mode flags missing-arg-type).

Net diff: 1 file changed, 7 insertions(+), 4 deletions(-).

### DIFF
--- a/beanie/odm/operators/find/comparison.py
+++ b/beanie/odm/operators/find/comparison.py
@@ -1,3 +1,6 @@
+from collections.abc import Mapping
+from typing import Any
+
 from beanie.odm.operators.find import BaseFindOperator
 
 
@@ -6,14 +9,14 @@ class BaseFindComparisonOperator(BaseFindOperator):
 
     def __init__(
         self,
-        field,
-        other,
+        field: Any,
+        other: Any,
     ) -> None:
         self.field = field
         self.other = other
 
     @property
-    def query(self):
+    def query(self) -> Mapping[str, Any]:
         return {self.field: {self.operator: self.other}}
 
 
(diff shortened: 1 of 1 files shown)


roman-right (approved):
LGTM. Same pattern as #1329 — annotating leftover untyped params on find comparison operators. Clean, no runtime change. Refs #1050.

Issue #1036, opened 2024-10-02
[BUG] When doing Annotated[str, Indexed(str, unique=True)] the uniqueness is not enforced


guyzmo:
**Describe the bug**

As described in [this issue](https://github.com/BeanieODM/beanie/issues/59#issuecomment-1565738344)
And discussed in [this issue](https://github.com/BeanieODM/beanie/issues/638)
And implemented in [this PR](https://github.com/BeanieODM/beanie/pull/762)
And according to [beanie's documentation](https://beanie-odm.dev/tutorial/indexes/#:~:text=For%20example%2C%20to%20create%20a%20unique%20index%3A) as well as [pymongo's documentation](https://pymongo.readthedocs.io/en/stable/api/pymongo/operations.html#pymongo.operations.IndexModel)

we can declare indexes with the following syntax:

```python
class User
    user_id: Annotated[str, Indexed(str)]
```

**To Reproduce**

But if we want to declare an index with the uniqueness property, it is not enforced:

```python
class User
    user_id: Annotated[str, Indexed(str, unique=True)]

user_1 = User(user_id: "foo").create()
user_2 = User(user_id: user_1.id).create()

assert len(User.find_all(User.user_id == user_1.user_id)) == 2
```

**Expected behavior**

There should be only one user with a given user_id, not two:

```python
class User
    user_id: Annotated[str, Indexed(str, unique=True)]

user_1 = User(user_id: "foo").create()
user_2 = User(user_id: user_1.id).create()

assert len(User.find_all(User.user_id == user_1.id)) == 1
```

**Additional context**

I'm still running an older version of beanie (v1.11.6), but it looks like this issue is still actual. [Cf discord](https://discord.com/channels/822196934973456394/1291031766676541481/1291031766676541481).

MrEarle:
When used with Annotated, you shouldn't provide the type to the Indexed function. Try this instead:

```py
class User
    user_id: Annotated[str, Indexed(unique=True)]
```

staticxterm:
Hi @MrEarle, thank you for the explanation. I just looked at the code as was having a hard time to understand why we can't also provide the type in this case also, but I see now that this is by design.
I believe this ticket should still stay open as at least we should clarify this usage in the [Indexes docs](https://beanie-odm.dev/tutorial/indexes/), to avoid any future confusion.

guyzmo:
ok, I did upgrade to latest, and that did solve the issue, so yes:

[...]

PR #1331, opened 2026-05-31
docs(indexes): clarify that `Annotated[T, Indexed(T, ...)]` is wrong

estelledc:
Refs #1036.

The reporter hit a real trap: \`Annotated[str, Indexed(str, unique=True)]\` *looks* like the \"explicit-type\" shape from the docs (\`Indexed(str, unique=True)\`), but in the \`Annotated\` position the first positional argument to \`Indexed()\` is treated as \`index_type\` (the pymongo index direction / kind), not as the field type. So \`unique=True\` doesn't apply the way the user expects.

@MrEarle explained the right shape in the issue thread:

> When used with \`Annotated\`, you shouldn't provide the type to the \`Indexed\` function. Try this instead:
> \`\`\`python
> user_id: Annotated[str, Indexed(unique=True)]
> \`\`\`

@staticxterm agreed the docs should call this out:

> I believe this ticket should still stay open as at least we should clarify this usage.

This PR adds a \`!!! warning\` callout under the existing \"Indexed in type annotation\" section with:

- The rule (don't pass the type inside \`Annotated[...]\`).
- A side-by-side correct/wrong snippet mirroring the failing example from #1036.
- A link back to the issue.

Pure docs change, 20 lines added.

### DIFF
--- a/docs/tutorial/indexes.md
+++ b/docs/tutorial/indexes.md
@@ -50,6 +50,26 @@ class Sample(Document):
     name: Indexed(str, unique=True)
 ```
 
+!!! warning "Don't pass the type when using `Annotated`"
+
+    The two syntaxes are not interchangeable. When wrapping with
+    `Annotated[...]`, do **not** pass the wrapped type to `Indexed()` —
+    in this position the first positional argument is interpreted as
+    `index_type` (the pymongo index direction / type), so passing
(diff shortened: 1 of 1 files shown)


roman-right (approved):
LGTM. Clear and correct warning — the side-by-side correct/wrong snippet is especially helpful for users hitting this trap. Refs #1036.

PR #1332, opened 2026-05-31
docs(tutorial): add Bulk operations page covering BulkWriter

### LINKED ISSUE #903: [BUG] Missing documentation for `beanie.odm.BulkWriter`
**Describe the bug**
As titled.
The documentation should have an example for `beanie.odm.BulkWriter` as well.

**To Reproduce**
See the current (v1.25.0) documentation.

**Expected behavior**
Have documentation on `beanie.odm.BulkWriter`.

estelledc:
Closes #903.

#903 flagged that \`BulkWriter\` had no tutorial coverage. This PR adds \`docs/tutorial/bulk-writer.md\` with:

- **Quick example** — \`async with BulkWriter() as bw\` plus \`Document.insert(bulk_writer=bw)\` showing the batch-then-commit shape that ships all queued operations in a single \`bulk_write\` on context exit.
- **Mixed-operation example** — insert + replace + update + delete all flushed in one round-trip.
- A list of every method that takes a \`bulk_writer\` keyword (\`insert\`, \`insert_many\`, \`replace\`, \`save\`, \`save_changes\`, \`update\`, \`set\`, \`inc\`, \`delete\`, plus the matching find/update/delete query-chain variants).
- **Manual-commit form** — \`bw = BulkWriter(); ...; await bw.commit()\` for when a context manager isn't a fit.
- **Constructor reference** — \`session\`, \`ordered\`, \`bypass_document_validation\`, \`comment\`, \`object_class\`.
- **Failure-mode note** — the \"All the operations should be for a same collection name\" \`ValueError\` and the \"one BulkWriter per collection\" rule it implies.

Wired into \`pydoc-markdown.yml\` between the existing \"Indexes\" and \"Multi-model pattern\" tutorial entries.

Pure docs change, 121 lines added across the new page + sidebar entry.

### DIFF
--- a/docs/tutorial/bulk-writer.md
+++ b/docs/tutorial/bulk-writer.md
@@ -1,0 +1,119 @@
+# Bulk operations with `BulkWriter`
+
+When you need to issue many writes to MongoDB at once, batching them into a
+single bulk request is dramatically faster than sending each one
+individually. Beanie exposes this through `BulkWriter`, an async context
+manager that collects operations and commits them in one round-trip when
+the block exits.
+
+## Quick example
+
+```python
+import asyncio
+
+from beanie import BulkWriter, Document, init_beanie
+from pymongo import AsyncMongoClient
+
+
+class Product(Document):
+    name: str
+    price: float
+
+
+async def main() -> None:
+    client = AsyncMongoClient("mongodb://localhost:27017")
+    await init_beanie(database=client.shop, document_models=[Product])
+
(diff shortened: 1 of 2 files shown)


roman-right (approved):
LGTM. Well-structured tutorial page covering all key BulkWriter use cases. One minor concern: the statement "on exception the queued operations are dropped intentionally" should be double-checked against BulkWriter.__aexit__ implementation for accuracy. Fixes #903.

PR #1343, opened 2026-07-30
Fix index direction float-to-int conversion for existing MongoDB indexes

gaurav-hub319:
## Summary
- Converts float index directions (`1.0` / `-1.0`) returned by MongoDB `index_information()` to ints before building `IndexModel`
- Prevents index comparison / construction failures when existing indexes use float direction values
- Same fix as discussed in https://github.com/BeanieODM/beanie/pull/811, rebased onto current `main` (`from_pymongo_index_information`)

## Test plan
- [ ] Create a document model with a simple ascending/descending index
- [ ] Ensure collection already has matching indexes whose directions are stored as floats
- [ ] Initialize Beanie / sync indexes and confirm no `IndexError` / type mismatch
- [ ] Confirm indexes continue to match and are not unnecessarily recreated

### DIFF
--- a/beanie/odm/fields.py
+++ b/beanie/odm/fields.py
@@ -623,12 +623,12 @@ class IndexModelField:
             if ("_id", 1) in fields:
                 continue
 
-            if (
-                fields
-                and isinstance(fields[0][1], float)
-                and (fields[0][1] == 1.0 or fields[0][1] == -1.0)
-            ):
-                fields = [(fields[0][0], int(fields[0][1]))]
+            fields = [
+                (field_name, int(direction))
+                if isinstance(direction, float) and direction in (1.0, -1.0)
+                else (field_name, direction)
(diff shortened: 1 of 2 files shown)


gaurav-hub319:
Added conversion for float to int for existing indexes while being fetched from mongo database due to pymongo check of indexes being 1,-1 or 2d

roman-right:
Thanks for this fix. I pushed a follow-up commit: the original fix only converted the direction on the first field of an index, so a compound index with a float direction on a later field would still fail. I extended it to convert every field's direction and added a test for that case, plus a test to confirm non-numeric directions like text indexes are left alone. Also rebased onto current main. Waiting on CI, will merge once green.

gaurav-hub319:
Thanks Roman for catching that edge case and adding the regression tests. CI is green now

PR #1345, opened 2026-08-02
Not: accept bool in expression type hint like the other logical ops do

### LINKED ISSUE #1000: [BUG] Logical operators are not type compliant when using bool operations

0x5t4l1n:
`Or`, `And`, and `Nor` already take `bool` in their `*expressions` hint so things like `Or(Product.price < 10, ...)` typecheck fine. `Not` was typed as `Mapping[str, Any]` only, so `Not(Product.price < 10)` got a mypy error about incompatible argument type.\n\nwidened `Not.__init__`'s parameter to `Mapping[str, Any] | bool` to match the pattern the other operators already use. fixes #1000.

### DIFF
--- a/beanie/odm/operators/find/logical.py
+++ b/beanie/odm/operators/find/logical.py
@@ -144,7 +144,7 @@ class Not(BaseFindLogicalOperator):
     <https://docs.mongodb.com/manual/reference/operator/query/not/>
     """
 
-    def __init__(self, expression: Mapping[str, Any]):
+    def __init__(self, expression: Mapping[str, Any] | bool):
         self.expression = expression
 
     @property


roman-right (approved):
Thanks, this matches the pattern the other logical operators already use. Merging.

PR #1346, opened 2026-08-03
fix(typing): preserve awaitable type for find-one upsert

### LINKED ISSUE #1342: [BUG] Pyright/Pylance reports `UpdateQuery` is not awaitable for the documented `find_one().upsert()` API
## Environment

- Beanie: 2.1.0
- Pyright: 1.1.411
- VS Code: 1.120.0
- Python: 3.12
- OS: macOS 26.5 arm64

## Description

Pyright/Pylance reports that the result of the documented
`find_one(...).upsert(...)` expression is not awaitable.

The code works correctly at runtime, and the same pattern is shown in the
official Beanie documentation:

```python
await Product.find_one(Product.name == "Tony's").upsert(
    Set({Product.price: 3.33}),
    on_insert=Product(
        name="Tony's",
        price=3.33,
        category=chocolate,
    ),
)
```

However, Pyright reports:

```text
"UpdateQuery" is not awaitable
"UpdateQuery" is incompatible with protocol "Awaitable[_T_co@Awaitable]"
"__await__" is not present

Pylance(reportGeneralTypeIssues)
```

## Minimal reproduction

```python
from beanie import Document


class Product(Document):
    name: str
    price: float


async def create_or_update_product() -> None:
    await Product.find_one(Product.name == "Tony's").upsert(
        Set({Product.price: 3.33}),
        on_insert=Product(
            name="Tony's",
            price=3.33,
        ),
    )
```

Run:

```bash
pyright example.py
```

## Actual result

Pyright reports the `await` expression as invalid because the inferred return
type of `upsert()` is `UpdateQuery`, which is not declared as awaitable.

```text
error: "UpdateQuery" is not awaitable
  "UpdateQuery" is incompatible with protocol "Awaitable[...]"
    "__await__" is not present
```

## Expected result

The documented expression should pass Pyright without requiring:

- `Any`;
- `cast(Awaitable[..., ...])`;
- a wrapper function;
- `# pyright: ignore[reportGeneralTypeIssues]`.

## Runtime behavior

The expression works correctly at runtime.

The concrete query returned for `find_one()` is an `UpdateOne`, and
`UpdateOne` implements `__await__`.

## Suspected cause

The return annotation of `upsert()` is widened to `UpdateQuery`:

```python
def upsert(...) -> "UpdateQuery":
    ...
    return self
```

The concrete returned object is an awaitable `UpdateOne`, but the declared
`UpdateQuery` base type does not expose `__await__`. Pyright therefore loses
the concrete awaitable type.

This appears to be an annotation issue rather than a runtime implementation
issue.

## Suggested fix

Preserve the concrete query type returned by `upsert()`, for example by using
`Self` where supported:

```python
from typing import Self

def upsert(...) -> Self:
    ...
    return self
```

Alternatively, annotate each concrete implementation with its actual
awaitable query type.

The same review may be useful for other fluent query methods that return
`self`, because widening them to `UpdateQuery` can discard capabilities
provided by concrete subclasses.

## Current workaround

A narrowly scoped suppression works:

```python
result = await Product.find_one(  # pyright: ignore[reportGeneralTypeIssues]
    Product.name == "Tony's"
).upsert(
    Set({Product.price: 3.33}),
    on_insert=Product(
        name="Tony's",
        price=3.33,
    ),
)
```

This avoids changing runtime code, but ideally the documented API should type
check without a suppression.

## Related issue

This is similar to the older Pylance/Pyright compatibility report:

- https://github.com/BeanieODM/beanie/issues/142
```

xianjianlf2:
Fixes #1342

`FindOne.update()` and `FindOne.upsert()` return the concrete `UpdateOne` query, whose awaitable behavior is available at runtime. The public annotations previously widened the result to `UpdateQuery`, so Pyright rejected valid `await Product.find_one(...).upsert(...)` code.

This keeps the concrete type through the fluent update methods and adds typing coverage for both calls.

Validation:
- pyright tests/typing/find.py
- ruff check and format check for changed files
- python3 -m compileall -q beanie tests/typing/find.py

### DIFF
--- a/beanie/odm/queries/find.py
+++ b/beanie/odm/queries/find.py
@@ -889,7 +889,7 @@ class FindOne(FindQuery[FindQueryResultType]):
         bulk_writer: BulkWriter | None = None,
         response_type: UpdateResponse | None = None,
         **pymongo_kwargs: Any,
-    ):
+    ) -> UpdateOne:
         """
         Create Update with modifications query
         and provide search criteria there
@@ -901,7 +901,8 @@ class FindOne(FindQuery[FindQueryResultType]):
         :return: UpdateMany query
         """
         self.set_session(session)
-        return (
+        return cast(
+            UpdateOne,
             self.UpdateQueryType(
                 document_model=self.document_model,
                 find_query=self.get_filter_query(),
@@ -912,7 +913,7 @@ class FindOne(FindQuery[FindQueryResultType]):
                 response_type=response_type,
                 **pymongo_kwargs,
             )
-            .set_session(session=self.session)
+            .set_session(session=self.session),
         )
 
     def upsert(
@@ -922,7 +923,7 @@ class FindOne(FindQuery[FindQueryResultType]):
         session: AsyncClientSession | None = None,
         response_type: UpdateResponse | None = None,
         **pymongo_kwargs: Any,
-    ):
+    ) -> UpdateOne:
         """
         Create Update with modifications query
         and provide search criteria there
@@ -935,7 +936,8 @@ class FindOne(FindQuery[FindQueryResultType]):
         :return: UpdateMany query
         """
         self.set_session(session)
-        return (
+        return cast(
+            UpdateOne,
             self.UpdateQueryType(
                 document_model=self.document_model,
                 find_query=self.get_filter_query(),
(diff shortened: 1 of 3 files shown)


roman-right (approved):
Thanks for the fix and for adding the typing test coverage. Merging.

